PR #8, opened 2022-01-13
Backport CONFIGURABLE_WASM_LIMITS from 2.1

swatanabe:


spoonincode:
Is it a requirement that this implementation match that as the 2.1 protocol feature? If not, it seems worthwhile to define a new protocol feature instead so that the fix for EOSIO/eos#10773 can be integrated

tbfleming:
> Is it a requirement that this implementation match that as the 2.1 protocol feature?

It's not

swatanabe:
Yeah, I'll add that fix.  There's another issue I'd like to fix as well (large custom sections can setcode but cannot execute), which wasn't really worth making a new protocol feature for.

tbfleming:
@swatanabe is it ok for me to merge this with my PR? I still need to add snapshot support and you've already ported most of what I need.

unittests/test-contracts/wasm_config_bios/wasm_config_bios.cpp
```diff
@@ -0,0 +1,35 @@
+#include <eosio/contract.hpp>
+
+extern "C" __attribute__((eosio_wasm_import)) void set_wasm_parameters_packed(const void*, std::size_t);
+#ifdef USE_EOSIO_CDT_1_7_X
```
tbfleming, line 4: Remove the ifdef; CDT 1.6.3 needs the imports

swatanabe:
> @swatanabe is it ok for me to merge this with my PR? I still need to add snapshot support and you've already ported most of what I need.

Yes.  I don't expect any further changes to conflict with what you're doing.

tbfleming (approved):

tbfleming:
While reviewing diffs for this I noticed that I missed bringing over the `entry_db` and `reset_memory_fail ` tests when I backported eos-vm. I'll create a PR for that this week.

PR #12, opened 2022-01-18
get_code_hash

tbfleming:


libraries/chain/include/eosio/chain/webassembly/interface.hpp
```diff
@@ -472,6 +472,26 @@ namespace webassembly {
...
+          *
+          * @ingroup authorization
+          * @param account - name of the account to check.
+          * @param code_sequence - Receives the number of times code has changed, or 0 if account does not exist.
+          * @param code_hash - Receives the code_hash, or empty if code not set or account does not exist.
+          * @param vm_type - Receives the vm type, or 0 if code not set or account does not exist.
+          * @param vm_version - Receives the vm version, or 0 if code not set or account does not exist.
+          *
+          * @return true if the account exists and has code.
+          * @return false otherwise.
```
swatanabe, line 486: Is the result bool necessary, given that it is strictly equivalent to `code_hash != sha256{}`?

libraries/chain/include/eosio/chain/webassembly/interface.hpp
```diff
@@ -472,6 +472,26 @@ namespace webassembly {
...
+         */
+         bool get_code_hash(
+            account_name account,
+            vm::argument_proxy<uint64_t*> code_sequence,
+            vm::argument_proxy<fc::sha256*> code_hash,
+            vm::argument_proxy<uint8_t*> vm_type,
+            vm::argument_proxy<uint8_t*> vm_version) const;
```
swatanabe, line 493: I would strongly prefer to have a single output buffer.

libraries/chain/apply_context.cpp
```diff
@@ -207,6 +207,26 @@ bool apply_context::is_account( const account_name& account )const {
    return nullptr != db.find<account_object,by_name>( account );
 }
 
+void apply_context::get_code_hash(
+   account_name account, uint64_t& code_sequence, fc::sha256& code_hash, uint8_t& vm_type, uint8_t& vm_version) const {
+
+   auto obj = db.find<account_metadata_object,by_name>(account);
+   if(!obj || obj->code_hash == fc::sha256{}) {
```
swatanabe, line 214: I think this could be simplified to just `if(!obj)` now.

libraries/chain/include/eosio/chain/webassembly/interface.hpp
```diff
@@ -508,6 +508,30 @@ namespace webassembly {
...
+          * struct {
+          *    uint64_t code_sequence;
+          *    fc::sha256 code_hash;
+          *    uint8_t vm_type;
+          *    uint8_t vm_version;
+          * } result;
+          *
+          * @ingroup authorization
+          * @param account - name of the account to check.
+          * @param struct_version - must be 0.
```
swatanabe, line 525: For other similar functions, we accept any value for the version, and and include the version in the result.  The rationale for this is that it allows the contract and the system to negotiate the highest version that they both understand.

swatanabe (approved):

libraries/chain/apply_context.cpp
tbfleming: The check is so if a new vm_type or vm_version are added in the future, these will still report 0 when there is no contract

swatanabe (approved):

PR #25, opened 2022-01-27
Replace N macro with operator ""_n

swatanabe:
Cherry picked from c5674d4c166639875f19cb7a2059941481ad9a66
with additional fixes.

Because the N macro did not have a project-specific prefix, it clashed
with its namesakes from third-party projects like LLVM. This commit
introduces a namespace-scoped operator ""_n and replaces all invocations
of the N macro with that operator.


tbfleming:
2.1's aggressive name checking was an incompatible change to the RPC API, so I left that part out of the back port. Its logic was also a bit clumsy, so I cleaned it up.

tbfleming (approved):

PR #45, opened 2022-02-26
Limit input transaction dedup time

heifner:
- Respect the `cpu-effort-percent` while removing expired transactions from deduplication list.

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -1597,7 +1597,8 @@ struct controller_impl {
                      uint16_t confirm_block_count,
                      const vector<digest_type>& new_protocol_feature_activations,
                      controller::block_status s,
-                     const std::optional<block_id_type>& producer_block_id )
+                     const std::optional<block_id_type>& producer_block_id,
+                     const fc::time_point& deadline )
    {
       EOS_ASSERT( !pending, block_validate_exception, "pending block already exists" );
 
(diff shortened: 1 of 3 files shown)


swatanabe:
I'm pretty sure that extra expired `transaction_objects` do not affect consensus.  Have you fully verified this?

If the timer expires, it will result in an empty block, right?

heifner:
Yes, an empty block if timer expires, similar to if it expires in the other steps of the producer_plugin start_block.
As for consensus, I was also a bit worried, but these are input transactions that have succeeded. Extra transactions in the list would result in a duplicate transaction exception instead of an expired transaction. Since either would only be during building of a block it should not affect consensus.

swatanabe (approved):
Also, just to confirm, this is not a backport of changes to EOSIO and there is no conversation elsewhere that needs to be referenced?

heifner:
This is not a backport, although EOSIO/eos could certainly use this change. I should push it up there as well.
The conversation was with Wax BPs on telegram.

heifner:
A bit of more info about this. WAX was seeing BPs taking more than `cpu-effort-percent` time in removing duplicate transactions. Upgrading to EOSIO/eos 2.1.x with your improved chainbase fixed the problem for them. Since Mandel also has the newer chainbase presumably that would also fix their issue.

heifner:
https://github.com/EOSIO/eos/pull/11071

PR #52, opened 2022-03-07
Add additional subjective cpu logging

heifner:
- Additional subjective cpu logging
- https://github.com/EOSIO/eos/pull/11074

### DIFF
--- a/plugins/producer_plugin/include/eosio/producer_plugin/subjective_billing.hpp
+++ b/plugins/producer_plugin/include/eosio/producer_plugin/subjective_billing.hpp
@@ -171,10 +171,12 @@ public:
       _block_subjective_bill_cache.clear();
    }
 
-   void on_block( const block_state_ptr& bsp, const fc::time_point& now ) {
-      if( bsp == nullptr ) return;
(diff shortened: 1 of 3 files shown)


plugins/producer_plugin/include/eosio/producer_plugin/subjective_billing.hpp
```diff
@@ -171,10 +171,12 @@ class subjective_billing {
       _block_subjective_bill_cache.clear();
    }
 
-   void on_block( const block_state_ptr& bsp, const fc::time_point& now ) {
-      if( bsp == nullptr ) return;
+   void on_block( fc::logger& log, const block_state_ptr& bsp, const fc::time_point& now ) {
+      if( bsp == nullptr || _disabled ) return;
```
swatanabe, line 175: Is there any way for this to cause a change in behavior?  It assumes that if `_disabled` is true, then there is nothing to remove, which should be valid as long as `_disabled` does not change.

heifner: `_disabled` is private and only can be turned to `true` via `disabled()`

swatanabe (approved):

PR #53, opened 2022-03-11
Track storage for multiple new features

brianjohnson5972:
Adding new container that wraps a multi-index and tracks the underlying storage over time.
Also provides ability to persist on shutdown.


heifner (changes requested):
As discussed, I think this would be better in fc. Also look at using cfile and cfile_datastream.

brianjohnson5972:
I think the failing docker test has to do with this branch being merged in from a forked repo (at least I remember having a similar issue with the eosio/eos repo. That coupled with moving everything to the fc module, I decided to close this and create #59 .

PR #56, opened 2022-03-13
net_plugin transaction dedup cache handling

heifner:
- Add option `p2p-dedup-cache-expire-time-sec` for configurating the transaction deduplication cache expiration time.
- Simplify transaction deduplication cache handling to just be a configurable expire time.
  - The dedup cache, `local_txns`, is simply an optimization in net_plugin to avoid tripping duplication transaction exception in `controller`.
  - The current implementation (before this PR) prevents clients from re-trying transactions as the `net_plugin` would not accept the transaction again until it expired or it appeared in a block.

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -2051,7 +2051,7 @@ namespace eosio {
 
    void dispatch_manager::bcast_transaction(const packed_transaction& trx) {
       std::shared_ptr<std::vector<char>> send_buffer;
-      auto now = fc::time_point::now();
+      const auto now = fc::time_point::now();
       for_each_connection( [this, &trx, &now, &send_buffer]( auto& cp ) {
          if( cp->is_blocks_only_connection() || !cp->current() ) {
             return true;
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -2071,17 +2029,13 @@ namespace eosio {
    }
 
    void dispatch_manager::bcast_transaction(const packed_transaction& trx) {
-      const auto& id = trx.id();
-      time_point_sec trx_expiration = trx.expiration();
-      node_transaction_state nts = {id, trx_expiration, 0, 0};
-
       std::shared_ptr<std::vector<char>> send_buffer;
-      for_each_connection( [this, &trx, &nts, &send_buffer]( auto& cp ) {
+      auto now = fc::time_point::now();
```
brianjohnson5972, line 2033: make this const to make it clear that now is the same for all connections.

brianjohnson5972, line 2033: make this const to make it clear that now is the same for all connections.

brianjohnson5972 (changes requested):
2 minor changes

heifner:
> 2 minor changes

I only see one requested change.

plugins/net_plugin/net_plugin.cpp
heifner: Done

brianjohnson5972 (approved):

PR #59, opened 2022-03-15
Track storage for multiple new features

brianjohnson5972:
Adding new container in fc submodule that wraps a multi-index and tracks the underlying storage over time.
Also provides ability to persist on shutdown.


plugins/chain_plugin/test/test_tracked_storage.cpp
```diff
@@ -0,0 +1,194 @@
+#define BOOST_TEST_MODULE tracked_storage
+#include <boost/test/included/unit_test.hpp>
+#include <eosio/chain_plugin/tracked_storage.hpp>
+#include <sstream>
+
+using boost::multi_index_container;
+using namespace boost::multi_index;
+using namespace eosio::chain_apis;
+
```
heifner, line 9: Same test that is in fc, right? Should remove from here.

heifner (changes requested):

plugins/chain_plugin/test/test_tracked_storage.cpp
brianjohnson5972: cleaned up now

heifner (approved):

PR #79, opened 2022-04-05
Transaction retry feature

heifner:
## Adds trx_retry_db which provides transaction retry capability to chain_plugin
- Changed `channels::transaction_ack`  from `std::pair<fc::exception_ptr, transaction_metadata_ptr>>` to `std::pair<fc::exception_ptr, packed_transaction_ptr>>` since `transaction_metadata` is not needed/used and this allows it to be called from `trx_retry_db` where `packed_transaction_ptr` is available but `transaction_metadata_ptr` is not readily available.

### New configuration args for nodoes
  -  `transaction-retry-max-storage-size-gb`
     - "Maximum size (in GiB) allowed to be allocated for the Transaction Retry feature. Setting above 0 enables this feature."
   - `transaction-retry-interval-sec` defaults to 20
      - "How often, in seconds, to resend an incoming transaction to network if not seen in a block."
   - `transaction-retry-max-expiration-sec` defaults to 90
      - "Maximum allowed transaction expiration for retry transactions, will retry transactions up to this value."
   - `transaction-finality-status-max-storage-size-gb`
      - "Maximum size (in GiB) allowed to be allocated for the Transaction Finality Status feature. Setting above 0 enables this feature."

### New command line options for cleos
- `--use-old-send-rpc` - "Use old RPC send_transaction, rather than new RPC /v2/chain/send_transaction"
- `-t,--return-failure-trace` - "Return partial traces on failed transactions"
   - This is ignored for now, feature to be added later in different PR
- `--retry` - "Request node to retry transaction until irreversible or in a block of height --retry-num-blocks"
  - Passes true to `send_transaction2_params` `retry_trx`
- `--retry-num-blocks` - "Request node to retry transaction until irreversible or in a block of height --retry-num-blocks"
  - Valid with the `--retry` option. If not specified then LIB is assumed for the retry transaction. Passed to `send_transaction2_params` `retry_trx_num_blocks`.
- `-o,--read-only` - "Specify a transaction is read-only"
  - Ignored for now, will be added in different PR. Calls `/v1/chain/compute_transaction` which is not implemented yet.

### New chain_api_plugin endpoint
- /v1/chain/send_transaction2  (send_transaction_params)
```
   struct send_transaction2_params {
      bool return_failure_traces = false; ///< Ignored. Will be enabled in future versions.
      bool retry_trx = false; ///< request transaction retry on validated transaction
      std::optional<uint16_t> retry_trx_num_blocks{}; ///< if retry_trx, report trace at specified blocks from executed or lib if not specified
      fc::variant transaction; ///< same format as send_transaction
   };
```

### Tests
- Unittests included.
- Integration test will be added in different PR.


plugins/chain_plugin/trx_retry_db.cpp
```diff
@@ -0,0 +1,326 @@
...
+   size_t size()const {
+      return _tracked_trxs.index().size();
+   }
+
+   void track_transaction( packed_transaction_ptr ptrx, std::optional<uint16_t> num_blocks, next_function<std::unique_ptr<fc::variant>> next ) {
+      EOS_ASSERT( _tracked_trxs.memory_size() < _max_mem_usage_size, tx_resource_exhaustion,
+                  "Transaction exceeded  transaction-retry-max-storage-size-gb limit: ${m} bytes", ("m", _tracked_trxs.memory_size()) );
+      auto i = _tracked_trxs.index().get<by_trx_id>().find( ptrx->id() );
+      if( i == _tracked_trxs.index().end() ) {
+         _tracked_trxs.insert( {std::move(ptrx), !num_blocks.has_value() ? lib_totem : *num_blocks, 0, {}, fc::time_point::now(), next} );
```
brianjohnson5972, line 117: Will this work correctly if user indicates num-blocks = 0?

plugins/chain_plugin/trx_retry_db.cpp
```diff
@@ -0,0 +1,326 @@
...
+
+   void ack_ready_trxs_by_block_num( uint32_t block_num ) {
+      const auto& idx = _tracked_trxs.index().get<by_ready_block_num>(); // less sort
+      // determine what to ack
+      std::vector<decltype(idx.begin())> to_process;
+      for( auto i = idx.begin(); i != idx.end(); ++i ) {
+         if( !i->is_ready() ) break;
+         if( i->num_blocks == lib_totem ) break;
+         // if we have reached requested block height then ack to user
+         if( i->ready_block_num() <= block_num ) {
```
brianjohnson5972, line 227: Couldn't we move this to the predicate for the for statement, i.e. can't we quick once we hit an iterator with ready_block_num() higher than block_num? And if that is correct, we can get rid of lines 224 and 225 as well.

plugins/chain_plugin/trx_retry_db.cpp
```diff
@@ -0,0 +1,326 @@
...
+         _tracked_trxs.erase( i->id() );
+      }
+   }
+
+   void ack_ready_trxs_by_lib( uint32_t lib_block_num ) {
+      const auto& idx = _tracked_trxs.index().get<by_block_num>();
+      // determine what to ack
+      std::vector<decltype(idx.begin())> to_process;
+      for( auto i = idx.lower_bound(1); i != idx.end(); ++i ) { // skip over not ready, block_num == 0
+         if( i->block_num > lib_block_num ) break;
```
brianjohnson5972, line 243: can we make an end iterator that is upper_bound(lib_block_num) and use that in the for predicate and drop this check?

brianjohnson5972 (changes requested):
Still have a little bit to review, but letting you get to work on these.

plugins/chain_plugin/trx_retry_db.cpp
```diff
@@ -0,0 +1,326 @@
...
+            break;
+         }
+         itr->next( std::static_pointer_cast<fc::exception>(
+               std::make_shared<expired_tx_exception>(
+                     FC_LOG_MESSAGE( error, "expired retry transaction ${id}, expiration ${e}, block time ${bt}",
+                                     ("id", itr->id())("e", itr->ptrx->expiration())
+                                     ("bt", block_timestamp) ) ) ) );
+         _tracked_trxs.erase( itr->id() );
+      }
+      return true;
```
brianjohnson5972, line 268: Can we drop the return true and just return void?

programs/cleos/main.cpp
```diff
@@ -224,6 +230,9 @@ void add_standard_transaction_options(CLI::App* cmd, string default_permission =
    cmd->add_option("--max-net-usage", tx_max_net_usage, localized("set an upper limit on the net usage budget, in bytes, for the transaction (defaults to 0 which means no limit)"));
 
    cmd->add_option("--delay-sec", delaysec, localized("set the delay_sec seconds, defaults to 0s"));
+   cmd->add_option("-t,--return-failure-trace", tx_rtn_failure_trace, localized("Return partial traces on failed transactions"));
+   cmd->add_option("--retry", tx_retry, localized("Request node to retry transaction until irreversible or in a block of height --retry-num-blocks"));
```
brianjohnson5972, line 234: What do you think about making this one --retry-irreversible and then making them exclusive? It will make the comments clearer. Otherwise the one below is confusing since you could never set --retry-num-blocks and ever end up getting an irreversible retry.

plugins/chain_plugin/trx_retry_db.cpp
```diff
@@ -0,0 +1,313 @@
...
+      return num_blocks == lib_totem;
+   }
+
+   bool is_ready()const {
+      return block_num != 0;
+   }
+
+   // for fc::tracked_storage, x3 for trx_meta as a very rough guess of trace variant size
+   // todo: add get_estimated_size to fc::variant for use here
+   size_t memory_size()const { return ptrx->get_estimated_size() * 3 + sizeof(*this); }
```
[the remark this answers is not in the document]
heifner: Updated comment.

plugins/chain_plugin/trx_retry_db.cpp
```diff
@@ -0,0 +1,313 @@
...
+         _tracked_trxs.erase( itr->id() );
+      }
+      return true;
+   }
+
+private:
+   const chain::controller& _controller; ///< the controller to read data from
+   chain::plugin_interface::compat::channels::transaction_ack::channel_type& _transaction_ack_channel;
+   const fc::microseconds _abi_serializer_max_time; ///< the maximum time to allow abi_serialization to run
+   size_t _max_mem_usage_size = 0;
```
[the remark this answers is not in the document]
heifner: Made them const

plugins/chain_plugin/trx_retry_db.cpp
```diff
@@ -0,0 +1,313 @@
...
+   }
+
+   void track_transaction( packed_transaction_ptr ptrx, std::optional<uint16_t> num_blocks, next_function<std::unique_ptr<fc::variant>> next ) {
+      EOS_ASSERT( _tracked_trxs.memory_size() < _max_mem_usage_size, tx_resource_exhaustion,
+                  "Transaction exceeded  transaction-retry-max-storage-size-gb limit: ${m} bytes", ("m", _tracked_trxs.memory_size()) );
+      auto i = _tracked_trxs.index().get<by_trx_id>().find( ptrx->id() );
+      if( i == _tracked_trxs.index().end() ) {
+         _tracked_trxs.insert( {std::move(ptrx), !num_blocks.has_value() ? lib_totem : *num_blocks, 0, {}, fc::time_point::now(), next} );
+      } else {
+         // already tracking trx_meta
```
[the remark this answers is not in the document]
heifner: Updated comment

plugins/chain_plugin/trx_retry_db.cpp
```diff
@@ -0,0 +1,313 @@
...
+            } catch( chain::abi_exception& ) {
+               tt.trx_trace_v = *trace;
+            }
+         } );
+      }
+   }
+
+   void on_block_start( uint32_t block_num ) {
+      // on forks rollback any accepted block transactions
+      rollback_to( block_num );
```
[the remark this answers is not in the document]
heifner: Clock has nothing to do with block_num. This rolls back according to block number

plugins/chain_plugin/trx_retry_db.cpp
```diff
@@ -0,0 +1,313 @@
...
+         // called on block_start, so any block_num greater or equal have been rolled back
+         if( i->block_num < block_num ) break;
+
+         to_process.emplace_back( _tracked_trxs.index().project<0>( i ) );
+      }
+      // perform rollback
+      for( auto& i : to_process ) {
+         _tracked_trxs.modify( i, [&]( tracked_transaction& tt ) {
+            // if forked out, then need to retry. Estimate last_try via block_num it was in.
+            tt.last_try = fc::time_point::now() - fc::microseconds( (i->block_num - tt.block_num) * chain::config::block_interval_us );
```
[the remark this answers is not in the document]
heifner: Modified `by_last_try`  to use a method that takes block_num into account. 
Thanks this is a lot cleaner.

plugins/chain_plugin/trx_retry_db.cpp
```diff
@@ -0,0 +1,313 @@
...
+            tt.block_num = 0;
+            tt.trx_trace_v.clear();
```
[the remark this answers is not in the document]
heifner: Large fork switch is executed on the main thread completely before anything else is allowed to execute. In normal cases you will get almost all the trxs again during that fork-switch. I don't see this as an issue. I recommend we leave it as is.

plugins/chain_plugin/trx_retry_db.cpp
heifner: It should, and it is a valid use case. A trx comes in early during speculative execution, is relayed to a bp which puts it in its current block and then sends you the block. You abort your block and process that trx in that block just received, you would then ack it instead of waiting for another block. Added a test case for it to the unittest.

plugins/chain_plugin/trx_retry_db.cpp
heifner: Yes. And can use upper_bound like you suggested below.

plugins/chain_plugin/trx_retry_db.cpp
heifner: Yes. done.

plugins/chain_plugin/trx_retry_db.cpp
heifner: Yes. Done.

programs/cleos/main.cpp
heifner: ok. Done.

plugins/chain_plugin/trx_retry_db.cpp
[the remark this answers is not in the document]
brianjohnson5972: I wasn't thinking. A speculative block will always be the next block number after head, rollbacks can only be when a produced block is received.

brianjohnson5972 (approved):

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -1122,6 +1113,28 @@ void chain_plugin::plugin_initialize(const variables_map& options) {
 
       my->chain.emplace( *my->chain_config, std::move(pfs), *chain_id );
 
+      if( options.count( "transaction-retry-max-storage-size-gb" )) {
+         EOS_ASSERT( options.at( "disable-api-persisted-trx" ).as<bool>(), plugin_config_exception,
+                     "disable-api-persisted-trx must be set to true for transaction retry feature" );
+         EOS_ASSERT( !options.count( "producer-name"), plugin_config_exception,
```
NatPDeveloper, line 1119: Is there a technical reason this cannot be run on a producer node? @heifner 

I have a use case where we run a blockchain locally for unit testing purposes and it would be too much to expect the user to spin up another API node just for this purpose.

heifner: The transaction retry feature re-sends a transaction to the network when it does not observe a successful locally speculatively executed transaction making it into a produced block. If you are running a single block production test network then the feature is not needed on the block producer because a transaction will either fail or succeed and be placed into a block.

PR #84, opened 2022-04-06
ENFS-43: Add Configured Whitelist to Ignore Number of Failures

oschwaldp-oci:
Do not apply the 3-strike rule to disable-subjective-account-billing accounts.

### DIFF
--- a/plugins/producer_plugin/include/eosio/producer_plugin/subjective_billing.hpp
+++ b/plugins/producer_plugin/include/eosio/producer_plugin/subjective_billing.hpp
@@ -114,7 +114,7 @@ public: // public for tests
 public:
    void disable() { _disabled = true; }
    void disable_account( chain::account_name a ) { _disabled_accounts.emplace( a ); }
-   bool is_account_disabled(const account_name& a ) { return _disabled_accounts.count( a ); }
+   bool is_account_disabled(const account_name& a ) const { return _disabled_accounts.count( a ); }
 
    /// @param in_pending_block pass true if pt's bill time is accounted for in the pending block
    void subjective_bill( const transaction_id_type& id, const fc::time_point& expire, const account_name& first_auth,
--- a/plugins/producer_plugin/include/eosio/producer_plugin/subjective_billing.hpp
+++ b/plugins/producer_plugin/include/eosio/producer_plugin/subjective_billing.hpp
@@ -114,6 +114,7 @@ public: // public for tests
 public:
    void disable() { _disabled = true; }
    void disable_account( chain::account_name a ) { _disabled_accounts.emplace( a ); }
+   bool is_account_disabled(const account_name & a ) { _disabled_accounts.count( a ); }
 
    /// @param in_pending_block pass true if pt's bill time is accounted for in the pending block
    void subjective_bill( const transaction_id_type& id, const fc::time_point& expire, const account_name& first_auth,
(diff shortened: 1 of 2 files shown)


plugins/producer_plugin/include/eosio/producer_plugin/subjective_billing.hpp
```diff
@@ -114,6 +114,7 @@ class subjective_billing {
 public:
    void disable() { _disabled = true; }
    void disable_account( chain::account_name a ) { _disabled_accounts.emplace( a ); }
+   bool is_account_disabled(const account_name & a ) { _disabled_accounts.count( a ); }
```
heifner, line 117: Prefer: `const account_name& a` vs` const account_name & a`.
Add `return`.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1912,7 +1912,8 @@ bool producer_plugin_impl::process_unapplied_trxs( const fc::time_point& deadlin
             auto trx_deadline = start + fc::milliseconds( _max_transaction_time_ms );
 
             auto first_auth = trx->packed_trx()->get_transaction().first_authorizer();
-            if( account_fails.failure_limit( first_auth ) ) {
+            auto max_fails_disabled = _subjective_billing.is_account_disabled( first_auth );
```
heifner, line 1915: Move this logic into `account_failures::failure_limit`. The `account_failures` constructor can take a reference to `_subjective_billing`. Then no need for `is_account_disabled` to be public.

heifner (changes requested):

plugins/producer_plugin/include/eosio/producer_plugin/subjective_billing.hpp
oschwaldp-oci: Done. Addressed in 60b4a62f0. Good catch.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1882,6 +1885,7 @@ class account_failures {
    };
 
    std::map<account_name, account_failure> failed_accounts;
+   eosio::subjective_billing subjective_billing;
```
heifner, line 1888: `const eosio::subjective_billing&` we don't want a copy.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1808,6 +1808,8 @@ class account_failures {
 public:
    constexpr static uint32_t max_failures_per_account = 3;
 
+   explicit account_failures( const eosio::subjective_billing& sb ) : subjective_billing(sb) {}
```
heifner, line 1811: Add comment that lifetime of `sb` must outlive `account_failures`

heifner (changes requested):

plugins/producer_plugin/producer_plugin.cpp
oschwaldp-oci: Yeah I thought that as soon as I pushed the change.  Good catch. Thanks.

plugins/producer_plugin/producer_plugin.cpp
oschwaldp-oci: Addressed in ac9936c

heifner (approved):

PR #86, opened 2022-04-07
Added capability for read-only txns / added compute_transaction to chain api

ndcgundlach:


heifner (changes requested):
Need to resolve conflicts, but looks good outside of that.

heifner (approved):

PR #93, opened 2022-04-12
Added compute_transaction integration test to verify read-only transaction

ndcgundlach:


tests/compute_transaction_test.py
```diff
@@ -0,0 +1,127 @@
...
+    txnData ='{"from":"account1","to":"account2","quantity":"5.0000 EOS","memo":"tx1"}'
+
+    Print("Sending read-only transfer")
+    trx = {
+        "actions": [{"account": "eosio.token", "name": "transfer",
+        "authorization": [{"actor": account2.name, "permission": "active"}],
+        "data": txnData}]
+    }
+
+    node.pushTransaction(trx, opts='--read-only', permissions=account2.name)
```
heifner, line 116: Should verify this returns True, otherwise it might have failed and your test would still pass.

heifner (changes requested):

ndcgundlach:
subsumed by https://github.com/eosnetworkfoundation/mandel/pull/108

PR #95, opened 2022-04-12
remove gettext/libintl dependency

spoonincode:
gettext is only used for cleos localization support. But, cleos has not had any effort to translate it thus far. Remove this unused dependency for now as it's an LGPL dep and may cause some grief depending on future platform support. When we get back around to localizing cleos we can reevaluate this need; it's possible something like boost::locale would be a better fit.

Port of EOSIO/eos#8118 & EOSIO/eos#10326

### DIFF
--- a/scripts/generate_bottle.sh
+++ b/scripts/generate_bottle.sh
@@ -45,7 +45,6 @@ echo "class Eosio < Formula
    option :universal
 
    depends_on \"gmp\"
-   depends_on \"gettext\"
    depends_on \"openssl@1.1\"
    depends_on \"libusb\"
    depends_on macos: :mojave
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,10 +10,6 @@ endif()
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/CMakeModules")
 
-if (UNIX AND APPLE)
-   list(APPEND CMAKE_PREFIX_PATH "/usr/local/opt/gettext")
-endif()
-
 include( GNUInstallDirs )
 include( InstallDirectoryPermissions )
 include( MASSigning )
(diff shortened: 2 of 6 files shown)


scripts/generate_bottle.sh
```diff
@@ -45,7 +45,6 @@ echo "class Eosio < Formula
    option :universal
 
    depends_on \"gmp\"
-   depends_on \"gettext\"
```
heifner, line 48: What about `gettext-devel` in`.cicd/platforms/centos-7.6-unpinned.dockerfile` and `.cicd/platforms/centos-7.6.dockerfile`. 

If `.cicd` is not going to be maintained it should be removed.

matthewdarwin:
I assume .cicd will come back at some point.

scripts/generate_bottle.sh
spoonincode: Yes absolutely -- anything not used or not tested ought to be removed. `.cicd`  & build scripts fall in that bucket now.

I've been reluctant to PR a change to zap all that since there really isn't a replacement for pinned builds outside the build script. But.. the build scripts are no longer tested or getting attention soooo :shrug:  :confused:

heifner: There is some thought that the `.cicd` will be used by a new cicd system, which is what Matthew Darwin is talking about below. Maybe go ahead and make the those two changes so we have them just in case.

spoonincode: imo, there is no viable path for `.cicd` as is: it's outdated, neglected, overcomplicated, interwoven with b1 infra, and of course currently completely untested. The best course of action appears to be to eliminate it and start anew with something else; using the old files as reference if need be.

heifner: I'll buy that.

heifner (approved):

PR #114, opened 2022-04-19
Don't apply subjective billing to read-only transactions

ndcgundlach:
This PR turns off subjective billing for read-only transactions.

*** Warning ***
An effect of this change is that compute_transaction becomes a potential DOS attack vector for any node that has enabled the compute_transaction end point.  Any exposed node should implement some form of throttling to prevent this sort of attack.

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -585,7 +585,7 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
                   }
                   exhausted = block_is_exhausted();
                } else {
-                   if (!trx->read_only)
+                   if (!disable_subjective_billing)
                       _subjective_billing.subjective_bill_failure( first_auth, trace->elapsed, fc::time_point::now() );
 
                   auto e_ptr = trace->except->dynamic_copy_exception();
@@ -599,7 +599,8 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
                   _unapplied_transactions.add_persisted( trx );
                } else {
                   // if db_read_mode SPECULATIVE then trx is in the pending block and not immediately reverted
-                  _subjective_billing.subjective_bill( trx->id(), expire, first_auth, trace->elapsed,
+                  if (!disable_subjective_billing)
+                     _subjective_billing.subjective_bill( trx->id(), expire, first_auth, trace->elapsed,
                                                        chain.get_read_mode() == chain::db_read_mode::SPECULATIVE );
                }
                send_response( trace );
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -561,7 +561,8 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
 
             bool disable_subjective_billing = ( _pending_block_mode == pending_block_mode::producing )
                                               || ( persist_until_expired && _disable_subjective_api_billing )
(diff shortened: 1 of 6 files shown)


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -584,7 +585,9 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
                   }
                   exhausted = block_is_exhausted();
                } else {
-                  _subjective_billing.subjective_bill_failure( first_auth, trace->elapsed, fc::time_point::now() );
+                   if (!trx->read_only)
+                      _subjective_billing.subjective_bill_failure( first_auth, trace->elapsed, fc::time_point::now() );
```
heifner, line 589: Also add `if` check around `subjective_bill` call below.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -584,7 +585,9 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+
                   auto e_ptr = trace->except->dynamic_copy_exception();
                   send_response( e_ptr );
                }
```
heifner, line 593: `persist_until_expired` should be false for read-only transactions. Change 
`app().get_method<incoming::methods::transaction_async>()(pretty_input, true, true,` call in chain_plugin to be
`app().get_method<incoming::methods::transaction_async>()(pretty_input, false, true,`

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -584,7 +585,9 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
                   }
                   exhausted = block_is_exhausted();
                } else {
-                  _subjective_billing.subjective_bill_failure( first_auth, trace->elapsed, fc::time_point::now() );
+                   if (!trx->read_only)
```
heifner, line 588: Actually, I think this is a bug in existing implementation. The `if` should be for `if ( ! disable_subjective_billing )`

tests/Node.py
```diff
@@ -851,6 +851,11 @@ def getEosBalances(self, accounts):
 
         return balances
 
+    # Gets accounts mapped to key. Returns json object
+    def getAccountSubjectiveInfo(self, account):
+        acct = self.getEosAccount(account)
+        return acct["subjective_cpu_bill_limit"]
```
heifner, line 857: Nice, forgot we already had that in 2.0.x

heifner (changes requested):

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -561,7 +561,8 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
 
             bool disable_subjective_billing = ( _pending_block_mode == pending_block_mode::producing )
                                               || ( persist_until_expired && _disable_subjective_api_billing )
-                                              || ( !persist_until_expired && _disable_subjective_p2p_billing );
+                                              || ( !persist_until_expired && _disable_subjective_p2p_billing )
+                                              || trx->read_only;
```
heifner, line 565: This logic needs to be applied to push_transaction near line 1960. There persist is `itr->trx_type == trx_enum_type::persisted`

heifner (changes requested):
Currently read-only transactions are signaled in controller as `accepted_transaction` and as `applied_transaction`. I think we should avoid signaling read-only transactions. Also read-only transaction traces are logged in deep-mind logger as applied. I think we should not deep-mind log them either.

PR #130, opened 2022-04-26
Fix pause timer calculation.

heifner:
Fix for the long-standing issue with not correctly accounting for WASM load times. We pause the billing timer when loading WASMs. However, we were not using the correct deadline after restarting the paused timer which sometimes would cause the transaction to fail even though it had not actually run its allotted time.

Previous code:
`deadline` was the user provided deadline which was either the transaction deadline calculated from `max-transaction-time` or the wall clock block deadline passed in from the `producer_plugin`. When setting the timer again, after pausing, it would restrict the run time to the deadline even if the deadline was the original transaction deadline as opposed to the block deadline. This would restrict the amount of time the transaction was allowed to run on "first call" (when WASM loaded). The transaction could then fail because it ran "too long" even though it had not actually executed for `max-transaction-time`.

Now:
Node configurable wall clock block deadline (configurable via `cpu-effort-percent`) and node configurable `max-transaction-time` are now provided to the transaction processor (`transaction_context`) so when pausing for WASM load times, the wall clock block deadline can be honored separately from the `max-transaction-time`.


unittests/api_tests.cpp
```diff
@@ -935,6 +946,54 @@ BOOST_AUTO_TEST_CASE(checktime_fail_tests) { try {
...
+   // 6ms should be lower than time to load wasm
+   auto before = fc::time_point::now();
+   BOOST_CHECK_EXCEPTION( call_test( t, test_pause_action<TEST_METHOD("test_checktime", "checktime_failure")>{},
+                                     0, 6, fc::raw::pack(10000000000000000000ULL), "pause"_n ),
+                          tx_cpu_usage_exceeded, is_tx_cpu_usage_exceeded );
+   auto after = fc::time_point::now();
+   // Test that it runs longer than specified limit of 6 to allow for wasm load time.
+   auto dur = (after - before).count();
+   dlog("elapsed ${e}us", ("e", dur) );
+   BOOST_CHECK( dur > 6500 );
```
swatanabe, line 989: Why 6500?  The 6ms passed to call_test is the explicit deadline, not the billing timer.  It should not be extended by pause.  The billing timer is 5ms.

unittests/api_tests.cpp
```diff
@@ -935,6 +946,54 @@ BOOST_AUTO_TEST_CASE(checktime_fail_tests) { try {
...
+
+   int64_t ram_bytes; int64_t net; int64_t cpu;
+   auto& rl = t.control->get_resource_limits_manager();
+   rl.get_account_limits( "pause"_n, ram_bytes, net, cpu );
+   BOOST_CHECK_EQUAL( cpu, -1 );
+   auto cpu_limit = rl.get_block_cpu_limit();
+   idump(("cpu_limit")(cpu_limit));
+   BOOST_CHECK( cpu_limit <= 250'000 );
+
+   // 6ms should be lower than time to load wasm
```
swatanabe, line 980: Having deadlines of 4ms, 5ms, and 6ms seems to make this test very dependent on the time required to compile wasm.  I'm seeing times under 6500 µs (i.e. failing the dur > 6500 check), so it wouldn't take a very much faster processor to invalidate the test results.

libraries/chain/transaction_context.cpp
```diff
@@ -431,14 +430,17 @@ namespace eosio { namespace chain {
...
+      total_paused_time += paused;
+
       pseudo_start = now - billed_time;
-      if( (pseudo_start + billing_timer_duration_limit) <= deadline ) {
-         _deadline = pseudo_start + billing_timer_duration_limit;
-         deadline_exception_code = billing_timer_exception_code;
-      } else {
-         _deadline = deadline;
-         deadline_exception_code = deadline_exception::code_value;
+      _deadline += total_paused_time;
```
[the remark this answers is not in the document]
heifner: No. The calls are guarded by `psuedo_start` being assigned `fc::time_point()` and `now - billed_time`.

unittests/api_tests.cpp
```diff
@@ -935,6 +946,54 @@ BOOST_AUTO_TEST_CASE(checktime_fail_tests) { try {
...
+   BOOST_CHECK_EXCEPTION( call_test( t, test_pause_action<TEST_METHOD("test_checktime", "checktime_failure")>{},
+                                     0, 6, fc::raw::pack(10000000000000000000ULL), "pause"_n ),
+                          tx_cpu_usage_exceeded, is_tx_cpu_usage_exceeded );
+   auto after = fc::time_point::now();
+   // Test that it runs longer than specified limit of 6 to allow for wasm load time.
+   auto dur = (after - before).count();
+   dlog("elapsed ${e}us", ("e", dur) );
+   BOOST_CHECK( dur > 6500 );
+
+   BOOST_CHECK_EXCEPTION( call_test( t, test_pause_action<TEST_METHOD("test_checktime", "checktime_failure")>{},
```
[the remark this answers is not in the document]
heifner: No this is a different account `pause` with a different action `test_pause_action`

unittests/api_tests.cpp
```diff
@@ -935,6 +946,54 @@ BOOST_AUTO_TEST_CASE(checktime_fail_tests) { try {
...
+   BOOST_CHECK_EQUAL( cpu, -1 );
+   auto cpu_limit = rl.get_block_cpu_limit();
+   idump(("cpu_limit")(cpu_limit));
+   BOOST_CHECK( cpu_limit <= 250'000 );
+
+   // 6ms should be lower than time to load wasm
+   auto before = fc::time_point::now();
+   BOOST_CHECK_EXCEPTION( call_test( t, test_pause_action<TEST_METHOD("test_checktime", "checktime_failure")>{},
+                                     0, 6, fc::raw::pack(10000000000000000000ULL), "pause"_n ),
+                          tx_cpu_usage_exceeded, is_tx_cpu_usage_exceeded );
```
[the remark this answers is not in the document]
heifner: `max_transaction_cpu_usage` is set to `4999` which is what is tripped here.

unittests/api_tests.cpp
heifner: WASM load times (on my machine) were taking about 10ms. `6500` is smaller than 10000 but higher than `4999`.

unittests/api_tests.cpp
heifner: Hard to argue with this assessment. I'm open for suggestions. Maybe add additional comments about this may fail on faster hardware. Could just remove the test now that I have proven to myself at least that this is working.

unittests/api_tests.cpp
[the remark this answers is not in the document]
swatanabe: But that timer should be paused, extending it beyond the 6ms limit.

libraries/chain/transaction_context.cpp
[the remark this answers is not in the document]
swatanabe: What I mean is, if a single transaction needs to compile more than one wasm, it look something like:
pause_resume 1:
```
total_paused_time = compile1
_deadline = original_deadline + compile1
```
pause/resume 2:
```
total_paused_time = compile1 + compile2
_deadline = original_deadline + compile1 + compile1 + compile2
```

because `total_paused _time` is already cumulative and we're accumulating again in `_deadline`.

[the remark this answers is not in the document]
heifner: ah, yes. Thanks!

swatanabe:
Extending the deadline from producer_plugin is not correct.  It's reasonable if the deadline is from --max-transaction-time, but not if it's because we need to ship out a block on time.

swatanabe:
`block_cpu_deadline` doesn't include the pauses from previous transactions does it?

heifner:
> `block_cpu_deadline` doesn't include the pauses from previous transactions does it?

It does not, it is the deadline calculated at the start of the transaction. We want to honor that deadline regardless how long we are paused. We want to make sure we ship a block out on time. The `block_cpu_usage_exceeded` will cause the producer_plugin to try the transaction in the next block.

swatanabe:
I agree that respecting `block_cpu_deadline` is necessary, but the block deadline coming from producer plugin is more restrictive and we're discarding some of that by using only `block_cpu_deadline`.  `block_cpu_deadline` is mostly objective, allowing only this transaction to decrease it subjectively.  `caller_deadline` is a wall-clock time limit and includes all time since the start of the block, billed or not.

heifner:
> I agree that respecting `block_cpu_deadline` is necessary, but the block deadline coming from producer plugin is more restrictive and we're discarding some of that by using only `block_cpu_deadline`. `block_cpu_deadline` is mostly objective, allowing only this transaction to decrease it subjectively. `caller_deadline` is a wall-clock time limit and includes all time since the start of the block, billed or not.

Often the producer plugin deadline is less restrictive because it is wall clock and not block cpu, but you are correct that it could be more restrictive, especially if cpu effort is set low. I was trying to avoid passing into push_transaction an indication if this was a block deadline.

swatanabe:
Summary of the problem as I understand it:
- transaction_checktime_timer manages two distinct deadlines, a wall clock deadline, and a billing clock deadline
- the limit set by the command line option `--max-transaction-time` is incorrectly treated as a wall-clock limit.

unittests/api_tests.cpp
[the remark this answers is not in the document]
heifner: This test is trying to show that a transaction gets its full 4999us to run even if paused for say 10ms. This 6 really could be 99999 which might make the test clearer. Any amount greater than 4999us is "ignored".

heifner:
> Summary of the problem as I understand it:
> 
> * transaction_checktime_timer manages two distinct deadlines, a wall clock deadline, and a billing clock deadline
> * the limit set by the command line option `--max-transaction-time` is incorrectly treated as a wall-clock limit.

Nicely stated. I'll look at this some more, but maybe the correct thing would be to provide push_transaction with the max-transaction-time and a block deadline and then let transaction_context do all this calculation. This would simplify the prdoducer_plugin and move all this into transaction_context.

libraries/chain/transaction_context.cpp
```diff
@@ -176,11 +178,22 @@ namespace eosio { namespace chain {
...
+      if( max_transaction_time_subjective != fc::microseconds::maximum() && (start + max_transaction_time_subjective) <= _deadline ) {
+         _deadline = start + (max_transaction_time_subjective == fc::microseconds::maximum() ?
+                              fc::hours(24*7*52) : max_transaction_time_subjective);
+         billing_timer_exception_code = tx_cpu_usage_exceeded::code_value;
+      }
+
+      // Possibly limit deadline to caller provided wall clock block deadline
+      if( block_deadline < _deadline ) {
+         _deadline = block_deadline;
+         billing_timer_exception_code = block_cpu_usage_exceeded::code_value;
```
swatanabe, line 191: Shouldn't `block_deadline` correspond to `deadline_exception`?  I thought `block_cpu_usage_exceeded` means that the billed time exceeds the objective block cpu limit.

libraries/chain/transaction_context.cpp
```diff
@@ -176,11 +178,22 @@ namespace eosio { namespace chain {
          billing_timer_exception_code = leeway_deadline_exception::code_value;
       }
 
-      billing_timer_duration_limit = _deadline - start;
+      // Possibly limit deadline to subjective max_transaction_time
+      if( max_transaction_time_subjective != fc::microseconds::maximum() && (start + max_transaction_time_subjective) <= _deadline ) {
+         _deadline = start + (max_transaction_time_subjective == fc::microseconds::maximum() ?
```
swatanabe, line 183: Checking for `fc::microseconds::maximum()` again is redundant.

libraries/chain/include/eosio/chain/transaction_context.hpp
```diff
@@ -161,14 +161,15 @@ namespace eosio { namespace chain {
 
          bool                          cpu_limit_due_to_greylist = false;
 
+         fc::microseconds              max_transaction_time_subjective;
+         fc::time_point                paused_time;
```
swatanabe, line 165: Do we need to have both `pseudo_start` and `paused_time`?

libraries/chain/transaction_context.cpp
```diff
@@ -176,11 +178,22 @@ namespace eosio { namespace chain {
...
+      if( block_deadline < _deadline ) {
+         _deadline = block_deadline;
+         billing_timer_exception_code = block_cpu_usage_exceeded::code_value;
+      }
 
-      // Check if deadline is limited by caller-set deadline (only change deadline if billed_cpu_time_us is not set)
-      if( explicit_billed_cpu_time || deadline < _deadline ) {
-         _deadline = deadline;
+      // Explicit billed_cpu_time_us should be used, block_deadline will be maximum unless in test code
+      if( explicit_billed_cpu_time ) {
```
swatanabe, line 195: The `deadline_exception` will be changed to a `block_cpu_usage_exceeded` if the timer is paused.  Maybe we should remove `deadline_exception`.

plugins/producer_plugin/include/eosio/producer_plugin/subjective_billing.hpp
```diff
@@ -114,7 +114,7 @@ class subjective_billing {
 public:
    void disable() { _disabled = true; }
    void disable_account( chain::account_name a ) { _disabled_accounts.emplace( a ); }
-   bool is_account_disabled(const account_name& a ) const { return _disabled_accounts.count( a ); }
+   bool is_account_disabled(const account_name& a ) const { return _disabled || _disabled_accounts.count( a ); }
```
swatanabe, line 117: Why is this change here?  It looks right, but seems unrelated.

unittests/api_tests.cpp
```diff
@@ -929,12 +947,62 @@ BOOST_AUTO_TEST_CASE(checktime_fail_tests) { try {
...
+   auto before = fc::time_point::now();
+   BOOST_CHECK_EXCEPTION( call_test( t, test_pause_action<TEST_METHOD("test_checktime", "checktime_failure")>{},
+                                     0, 9999, 500, fc::raw::pack(10000000000000000000ULL), "pause"_n ),
+                          tx_cpu_usage_exceeded, is_tx_cpu_usage_exceeded );
+   auto after = fc::time_point::now();
+   // Test that it runs longer than specified limit of 24'999 to allow for wasm load time.
+   auto dur = (after - before).count();
+   dlog("elapsed ${e}us", ("e", dur) );
+   BOOST_CHECK( dur >= 24'999 ); // should never fail
+   BOOST_CHECK( dur > 25'500 );  // if this fails but duration is >= 24'999, then could just be faster wasm load times
```
swatanabe, line 998: So, this test will pass iff loading the wasm takes at least 1.5 ms, which I think is safe for now, but should be called out explicitly.

unittests/api_tests.cpp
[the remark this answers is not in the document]
swatanabe: If it's cached, then there's no pause, so I'm not sure what this is supposed to be testing.

unittests/api_tests.cpp
```diff
@@ -929,12 +947,62 @@ BOOST_AUTO_TEST_CASE(checktime_fail_tests) { try {
...
+   int64_t ram_bytes; int64_t net; int64_t cpu;
+   auto& rl = t.control->get_resource_limits_manager();
+   rl.get_account_limits( "pause"_n, ram_bytes, net, cpu );
+   BOOST_CHECK_EQUAL( cpu, -1 );
+   auto cpu_limit = rl.get_block_cpu_limit();
+   idump(("cpu_limit")(cpu_limit));
+   BOOST_CHECK( cpu_limit <= 150'000 );
+
+   // Verify that the restriction on the transaction of 24'999 is honored even though there is wall clock time to
+   // load the wasm. If this test fails it is possible that the wasm loaded faster or slower than expected.
```
swatanabe, line 988: I think we need two tests
- The 24999 limit *is* extended by a pause (this check)
- The block deadline *is not* extended by a pause (I don't see this)

swatanabe: Oh, and one more:
- The cpu limit is is originally before the block deadline, but is extended past it by a pause.

unittests/api_tests.cpp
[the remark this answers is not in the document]
heifner: It tests that it is throwing the correct exception for hitting max_transaction_time. Added a comment and added another test case to verify correct exception for hitting block deadline.

libraries/chain/transaction_context.cpp
heifner: I was thinking the user would not care the difference. But our exception does say `billable`, so I'll go back to having deadline for this. It probably is useful to have diff exceptions for these two cases.

libraries/chain/transaction_context.cpp
heifner: Removed

libraries/chain/transaction_context.cpp
heifner: Actually, I think having diff exceptions is likely useful.

plugins/producer_plugin/include/eosio/producer_plugin/subjective_billing.hpp
heifner: It is unrelated. I will pull it out into a diff PR.

unittests/api_tests.cpp
heifner: Updated comment.

unittests/api_tests.cpp
heifner: Added tests.

unittests/api_tests.cpp
swatanabe: So, I clearly don't know how to do subtraction...

unittests/api_tests.cpp
```diff
@@ -995,12 +997,111 @@ BOOST_AUTO_TEST_CASE(checktime_pause_test) { try {
...
+                          deadline_exception, is_deadline_exception );
+   auto after = fc::time_point::now();
+   // Test that it runs longer than specified limit of 24'999 to allow for wasm load time.
+   auto dur = (after - before).count();
+   dlog("elapsed ${e}us", ("e", dur) );
+   BOOST_CHECK( dur >= 25'000 ); // should never fail
+   // If this check fails but duration is >= 25'000 (previous check did not fail), then the check here is likely
+   // because it took longer than 2.5 ms for checktime to trigger, trace to be created, and to get to the now() call.
+   // Timer is paused during WASM load, so checktime will not trigger until after WASM is loaded.
+   BOOST_CHECK( dur < 150'000 ); // should not run to max_transaction_time
```
swatanabe, line 1103: This seems too long and does not match the comment of 2.5 ms.  For this test, I would expect to see:
- The base timeout (25 ms) is an upper bound on the wasm load time
- The expected duration is an upper bound on timeout + checktime overhead
- The expected duration is a lower bound on timeout + WASM compile time (not satisfied)

A high value here means that this test will always pass even if the block deadline is extended.

swatanabe: If random noise makes this test to flaky, then we might have to give up on testing this condition.

heifner: I separated this test into two. One which gives time for the WASM to load but is limited to block deadline. This test I tweaked the values and comments to more closely reflect this particular test. Turns out on my machine I was seeing WASM load times of 35ms.

swatanabe (approved):

unittests/api_tests.cpp
```diff
@@ -1000,8 +1000,8 @@ BOOST_AUTO_TEST_CASE(checktime_pause_max_trx_cpu_extended_test) { try {
    // This assumes that loading the WASM takes at least 1.5 ms
    // If this check fails but duration is >= 24'999 (previous check did not fail), then the check here is likely
    // because WASM took less than 1.5 ms to load.
-   BOOST_CHECK( dur > 26'500 );
-   BOOST_CHECK( dur < 150'000 ); // Should not run to block_cpu_usage deadline
+   BOOST_CHECK_MESSAGE( dur > 26'500, "elapsed " << std::to_string(dur) << "us" );
```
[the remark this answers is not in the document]
heifner: `api_tests.cpp(1102): error: in "api_tests/checktime_pause_block_deadline_not_extended_test": check dur < 30'000 has failed` It wasn't for me.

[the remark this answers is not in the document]
swatanabe: Odd.  That's what I'd expect from `BOOST_CHECK`.  `BOOST_TEST` is supposed to decompose the expression and has been the recommended macro to use for a long time.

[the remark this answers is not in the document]
heifner: Maybe you meant the `to_string` is not needed. Removed the `to_string` calls.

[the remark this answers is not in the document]
swatanabe:
```
mandel/unittests/api_tests.cpp(1003): error: in "api_tests/checktime_pause_max_trx_cpu_extended_test": check dur < 26'500 has failed [28130 >= 26500]
```

swatanabe (approved):

unittests/api_tests.cpp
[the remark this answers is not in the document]
swatanabe: `BOOST_TEST( dur < 26'500 );`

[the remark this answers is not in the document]
heifner: How did you get it to print that? Some special command line arg to unit_test?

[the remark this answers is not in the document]
heifner: Oh, you said `BOOST_TEST`. Ah, need to switch to using that in the future. Thanks.

PR #131, opened 2022-04-26
Fix strict aliasing violations in database_utils.hpp

spoonincode:
This ports over EOSIO/eos#9761

> There are a number of uses of reinterpret_cast in database_utils.hpp. They violate the strict aliasing rule.
The fix is using memcpy to assign values between non-compatible types.

### DIFF
--- a/libraries/chain/include/eosio/chain/database_utils.hpp
+++ b/libraries/chain/include/eosio/chain/database_utils.hpp
@@ -107,34 +107,60 @@ namespace fc {
    }
 
    inline
+   void float64_to_double (const float64_t& f, double& d) {
+      memcpy(&d, &f, sizeof(d));
+   }
+
+   inline
+   void double_to_float64 (const double& d, float64_t& f) {
+      memcpy(&f, &d, sizeof(f));
+   }
+
+   inline
+   void float128_to_uint128 (const float128_t& f, eosio::chain::uint128_t& u) {
+      memcpy(&u, &f, sizeof(u));
+   }
+
+   inline
+   void uint128_to_float128 (const eosio::chain::uint128_t& u,  float128_t& f) {
+      memcpy(&f, &u, sizeof(f));
+   }
+
+   inline
    void to_variant( const float64_t& f, variant& v ) {
-      v = variant(*reinterpret_cast<const double*>(&f));
+      double double_f;
+      float64_to_double(f, double_f);
+      v = variant(double_f);
    }
 
    inline
    void from_variant( const variant& v, float64_t& f ) {
-      from_variant(v, *reinterpret_cast<double*>(&f));
+      double double_f;
+      from_variant(v, double_f);
+      double_to_float64(double_f, f);
    }
 
    inline
    void to_variant( const float128_t& f, variant& v ) {
       // Assumes platform is little endian and hex representation of 128-bit integer is in little endian order.	
-      const eosio::chain::uint128_t as_bytes = *reinterpret_cast<const eosio::chain::uint128_t*>(&f);
+      char as_bytes[sizeof(eosio::chain::uint128_t)];
+      memcpy(as_bytes, &f, sizeof(as_bytes));
       std::string s = "0x";	
-      s.append( to_hex( reinterpret_cast<const char*>(&as_bytes), sizeof(as_bytes) ) );	
+      s.append( to_hex( as_bytes, sizeof(as_bytes) ) );
       v = s;
    }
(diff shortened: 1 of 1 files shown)


libraries/chain/include/eosio/chain/database_utils.hpp
```diff
@@ -106,35 +106,61 @@ namespace fc {
...
    }
 
    inline
    void to_variant( const float128_t& f, variant& v ) {
       // Assumes platform is little endian and hex representation of 128-bit integer is in little endian order.	
-      const eosio::chain::uint128_t as_bytes = *reinterpret_cast<const eosio::chain::uint128_t*>(&f);
+      char as_bytes[sizeof(eosio::chain::uint128_t)];
+      memcpy(as_bytes, &f, sizeof(as_bytes));
       std::string s = "0x";	
-      s.append( to_hex( reinterpret_cast<const char*>(&as_bytes), sizeof(as_bytes) ) );
```
swatanabe, line 124: Cast to `const char*` is legal and doesn't require memcpy.  This could just drop `as_bytes` and cast `&f`.

libraries/chain/include/eosio/chain/database_utils.hpp
```diff
@@ -106,35 +106,61 @@ namespace fc {
       from_variant(v, oid._id);
    }
 
+   inline
+   void float64_to_double (const float64_t& f, double& d) {
```
swatanabe, line 110: softfloat.hpp has has `double from_softfloat64(float64_t)` and `float64_t to_softfloat64(double)`.  I would use those and define the float128 versions similarly.  (Note that C++20 has std::bit_cast)

libraries/chain/include/eosio/chain/database_utils.hpp
spoonincode: Are you suggesting to change it? I'm mostly indifferent, as it was already approved previously like this.

libraries/chain/include/eosio/chain/database_utils.hpp
spoonincode: At first I wasn't sure if eos-vm considered these softfloat functions part of its public interface. But it seems they're present even without `ENABLE_SOFTFLOAT` so maybe it does. The code reuse would probably be a good thing, I suppose, even in this trivial case.

swatanabe: They're not from eos-vm.  They're from berkeley softfloat.

libraries/chain/include/eosio/chain/database_utils.hpp
swatanabe: Up to you.  It's not a big deal, and the code works either way.

libraries/chain/include/eosio/chain/database_utils.hpp
spoonincode: Well that's the thing, we bring in softfloat two places: eos-vm and as a library in eosio.

eos-vm exposes the softfloat headers in a way that suggests they're public interface. (it's included in a cmake PUBLIC interface -- which they probably need to be due to being a header only library -- but they're not in an /impl/ or /detail/ directory and/or namespace either)

The library that's pulled for the rest of the eosio build is primarily used for softfloat injection in EOS VM OC. But mid/long term I expect that injection to go away. There are a few very very minor usages of softfloat outside of EOS VM OC injection -- funny enough already in `database_utils.hpp` does it include `softfloat.hpp`. Kinda lousy to double down depending on this library for just a few lines of code here and there; even more lousy if we need to go add float128 versions to said library :confused:

swatanabe: We already have a direct dependency here in the types `float32_t` and `float64_t`

swatanabe: If we do eliminate those types, then the dependency on softfloat sill naturally go away.  Anyway, I'm approving this PR as-is.  Feel free to take or ignore my suggestions at your own discretion.

swatanabe (approved):

libraries/chain/include/eosio/chain/database_utils.hpp
spoonincode: I'll go ahead and take as-is since 1) it was approved in the past and 2) there doesn't seem to be anything straight out incorrect with the changes. You can submit a PR with your improvements if you have a chance.

PR #137, opened 2022-04-26
remove build scripts & b1's cicd

spoonincode:
Remove the entirety of build scripts & b1's CICD. Both of these are untested in Mandel's repo, are rotting away (already incorrect in places), and are unmaintained (none of the Mandel developers use the scripts).

There is no immediate replacement for any of the removed functionality beyond what is already present. Probably the most significant loss is the pinned build script as that was still somewhat arguably user facing and offered some utility that has no current substitute. The desire is to have a reasonable replacement before 3.1rc1 is ready.

I see no real downside with removing all this now, with the slight annoyance of a break in git history for anything restored later (something like the package creation scripts could see a revival in some form or another).


spoonincode:
I guess the way I see it is that 3.0/3.0.5 have _already_ officially removed that feature. Yeah, it's a super important feature, but by ignoring it, not wiring it up for testing, and then cutting multiple RC releases without it, the project has implicitly officially removed the feature. It's gone.

Keeping the files around doesn't bring back the feature. Instead, leaving rubbish around seems like a disaster waiting to happen; it sets a trap.

If we feel the feature is important, we can add it back in a supported way. Would an issue as simple as 'bring back pinned builds & binaries' work?

heifner:
> Would an issue as simple as 'bring back pinned builds & binaries' work?

You make a good point. Yes, please add an issue and link it to this.

PR #144, opened 2022-04-27
Add terminate-at-block option to nodeos

spoonincode:
Ports over EOSIO/eos#8220, EOSIO/eos#10676, EOSIO/eos#10735 and EOSIO/eos#10892

> Add `--terminate-at-block` command line option to `nodeos`. Allows pre-determined shutdown during replay especially, but works while syncing and producing as well.


heifner (commented):
All the long running tests are wired up, unless you added one I didn't see. It is not too bad to add new ones. I think we need to wire them up as part of this PR.

PR #146, opened 2022-04-27
refactor release-build-test to eliminate sole usage of jq

spoonincode:
Refactor out the sole usage of `jq`. Python is already a requirement to run tests so lean on that instead. Removes `jq` from being a requirement.

### DIFF
--- a/tests/release-build.sh
+++ b/tests/release-build.sh
@@ -10,49 +10,13 @@ echo 'asserts. This test checks that debug flag. Anyone intending to build and i
 echo 'nodeos from source should perform a "release build" which excludes asserts and'
 echo 'debugging symbols, and performs compiler optimizations.'
 echo ''
-# check for jq
-if ! $(jq --version 1>/dev/null); then
-    echo 'ERROR: Test requires jq, but jq was not found in your PATH!'
-    echo ''
-    exit 1
-fi
-# find nodeos
-[[ $(git --version) ]] && cd "$(git rev-parse --show-toplevel)/build/programs/nodeos" || cd "$(dirname "${BASH_SOURCE[0]}")/../programs/nodeos"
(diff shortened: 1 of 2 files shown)


tests/release-build.sh
```diff
@@ -10,49 +10,13 @@ echo 'asserts. This test checks that debug flag. Anyone intending to build and i
...
 echo 'debugging symbols, and performs compiler optimizations.'
 echo ''
-# check for jq
-if ! $(jq --version 1>/dev/null); then
-    echo 'ERROR: Test requires jq, but jq was not found in your PATH!'
-    echo ''
-    exit 1
-fi
-# find nodeos
-[[ $(git --version) ]] && cd "$(git rev-parse --show-toplevel)/build/programs/nodeos" || cd "$(dirname "${BASH_SOURCE[0]}")/../programs/nodeos"
```
swatanabe, line 20: I don't see why this is being removed.  It doesn't depend on `jq`.

spoonincode: Running test scripts in rando directories is not something that is universally supported throughout our tests so I don't see the need to copy and paste gobbledygook like this all over a limited set of tests. Example:
```sh
$ pwd
/home/spoon/f/mandel/build.release/tests
$ ./nodeos_protocol_feature_test.py
2022-04-28T17:51:32.427614  BEGIN
2022-04-28T17:51:32.427940  Test failed.
2022-04-28T17:51:32.428014  Shut down the cluster.
(snip)
FileNotFoundError: [Errno 2] No such file or directory: 'programs/cleos/cleos'
```

I thought the PR had enough of an undertone that its purpose was to remove complexity, so I didn't explicitly call this change out.

swatanabe (approved):

PR #153, opened 2022-04-27
Better error message on failure to nodeos and keosd from cleos

heifner:
Rework of https://github.com/EOSIO/eos/pull/11046

Example error output:

```
Failed http request to nodeos at http://127.0.0.1:8888/; is nodeos running?
  Common issue is message size too large. Check the log of nodeos.
  Error: connect: Connection refused

```


programs/cleos/main.cpp
```diff
@@ -254,10 +254,16 @@ fc::variant call( const std::string& url,
...
-      else if(url == ::wallet_url)
-         std::cerr << localized("Failed to connect to ${k} at ${u}; is ${k} running?", ("k", key_store_executable_name)("u", url)) << std::endl;
+      std::string exec_name;
+      if(url == ::url) {
+         exec_name = node_executable_name;
+      } else if(url == ::wallet_url) {
+         exec_name = key_store_executable_name;
+      }
+      std::cerr << localized( "Failed http call to ${n} at ${u}; is ${n} running?\n"
+                              "  Common issue is message size too large. Check the log of ${n}.\n"
```
swatanabe, line 264: I would prefer the term "http request."  Also, I think it should be possible to distinguish between failing to establish a connection and having the connection terminated and issue appropriate advise for each case.

swatanabe (approved):

PR #160, opened 2022-04-29
increase HTTP server's default maximum POST size to 2MB

spoonincode:
I'd say at least once a month I see a user ensnared by the HTTP server's default maximum POST size when trying to set a "large" (512KB-1MB) contract on their account. There's even a recent open issue in b1's repo with someone who fell victim to this trap (EOSIO/eos#11085)

The reason this problem creeps up is because the WASM in the request sent to the HTTP server in keosd is encoded as a hex byte array, effectively doubling its size in the POST payload.

#153 improves the error message in this situation. But I really can't see any downside to just increasing the default limit to 2MB. This allows users to set contracts up to just under 1MB in size -- the maximum size cleos can set contract code to as-is anyways. A 2MB instead of 1MB limit would even seem unlikely to impact a nodeos RPC node exposed publicly with no proxy protection in front of it.

### DIFF
--- a/plugins/http_plugin/http_plugin.cpp
+++ b/plugins/http_plugin/http_plugin.cpp
@@ -193,7 +193,7 @@ class http_plugin_impl : public std::enable_shared_from_this<http_plugin_impl> {
          string                         access_control_allow_headers;
          string                         access_control_max_age;
          bool                           access_control_allow_credentials = false;
-         size_t                         max_body_size{1024*1024};
+         size_t                         max_body_size{2*1024*1024};
 
          websocket_server_type    server;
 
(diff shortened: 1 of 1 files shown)


spoonincode:
something bad with that test on main from which this was branched. fix is over in #144's ba933de so once that PR goes in I'll merge in the fix here

poplexity (commented):
Does keosd also use the http plugin?  Only thing worse than understanding the unclear error message is finally figuring it out and updating the setting on nodeos just to keep getting the same error message after updating nodeos because keosd is still failing.

heifner (approved):

heifner:
> Does keosd also use the http plugin? Only thing worse than understanding the unclear error message is finally figuring it out and updating the setting on nodeos just to keep getting the same error message after updating nodeos because keosd is still failing.

Yes, keosd uses the same http plugin.

PR #163, opened 2022-04-29
Added checks for special cases where to not apply txn deadline

ndcgundlach:


plugins/producer_plugin/include/eosio/producer_plugin/subjective_billing.hpp
```diff
@@ -113,6 +113,7 @@ class subjective_billing {
 
 public:
    void disable() { _disabled = true; }
+   bool is_disabled() const {return _disabled;}
    void disable_account( chain::account_name a ) { _disabled_accounts.emplace( a ); }
    bool is_account_disabled(const account_name& a ) const { return _disabled_accounts.count( a ); }
```
heifner, line 118: I think you should check, `if _disabled` in this method and just call it instead of adding `is_disabled()`

heifner (changes requested):

heifner:
I needed this change for https://github.com/eosnetworkfoundation/mandel/pull/130.
It will be included in that PR.

Issue #58, opened 2022-03-14
backport: p2p timeout  from 2.1


matthewdarwin:
**EOSUSA Michael, [14.03.22 10:48]
i think there were 2 parts... one is the p2p issue where it would get all peers unlinkable....  did the fixes also include anything related to the original issue where the p2p sessions hang open for 15min (which is what triggers all those unlinkable blocks in the first place)

Kevin Heifner, [14.03.22 10:49]
The timeout feature is in 2.1, it will need to be back-ported to Mandel**

heifner:
Included as part of https://github.com/eosnetworkfoundation/mandel/pull/169

PR #169, opened 2022-05-01
Backport eosio/eos 2.1 version of net_plugin

heifner:
- Includes special handling of protocol version to allow integration with 2.1 eosio/eos versions of nodeos.
- Includes `--p2p-keepalive-interval-ms` see https://github.com/EOSIO/eos/pull/9694
- Resolves: https://github.com/eosnetworkfoundation/mandel/issues/58


spoonincode:
I'm not really sure how to approach reviewing this. I originally started looking it as a diff from 2.1 vs Mandel and then backing out changes from that diff that are included in mandel via other PRs (like #56) but this was still unwieldy to really grok. I've at least run a node connected to some mainnet peers and all seems well. Anything specific that would be interesting to double check?

heifner:
> I've at least run a node connected to some mainnet peers and all seems well. Anything specific that would be interesting to double check?

That is the main thing. This PR includes special handling to "lie" to `eosio/eos 2.1` nodes telling them it is an `eosio/eos 2.0` node because Mandel does not understand `eosio/eos 2.1` `packed_transaction` or `signed_block`. We have many integration tests that test Mandel <-> Mandel. But no multi-version tests at this time. I Also verified manually it would sync from a 2.1 and that a 2.1 would sync from it.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1238,67 +1301,128 @@ namespace eosio {
...
+         const uint32_t payload_size = which_size + fc::raw::pack_size( v );
 
-      auto send_buffer = std::make_shared<vector<char>>( buffer_size );
-      fc::datastream<char*> ds( send_buffer->data(), buffer_size );
-      ds.write( header, header_size );
-      fc::raw::pack( ds, unsigned_int( which ) );
-      fc::raw::pack( ds, v );
+         const char* const header = reinterpret_cast<const char* const>(&payload_size); // avoid variable size encoding of uint32_t
+         constexpr size_t header_size = sizeof( payload_size );
+         static_assert( header_size == message_header_size, "invalid message_header_size" );
```
jgiszczak, line 1346: As above.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1238,67 +1301,128 @@ namespace eosio {
...
+      static send_buffer_type create_send_buffer( const net_message& m ) {
+         const uint32_t payload_size = fc::raw::pack_size( m );
 
-      auto send_buffer = std::make_shared<vector<char>>(buffer_size);
-      fc::datastream<char*> ds( send_buffer->data(), buffer_size);
-      ds.write( header, header_size );
-      fc::raw::pack( ds, m );
+         const char* const header = reinterpret_cast<const char* const>(&payload_size); // avoid variable size encoding of uint32_t
+         constexpr size_t header_size = sizeof(payload_size);
+         static_assert( header_size == message_header_size, "invalid message_header_size" );
```
jgiszczak, line 1327: This static_assert is fairly useless. It carefully verifies that a uint32_t is 4 bytes. So, it complies with the C++11 standard? How helpful. If we really need to check for such a thing, we should put it in cmake.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1238,67 +1301,128 @@ namespace eosio {
...
+            send_buffer = create_send_buffer( sb );
+         }
+         return send_buffer;
+      }
+
+   private:
+
+      static std::shared_ptr<std::vector<char>> create_send_buffer( const signed_block_ptr& sb ) {
+         static_assert( signed_block_which == fc::get_index<net_message, signed_block>() );
+         // this implementation is to avoid copy of signed_block to net_message
```
jgiszczak, line 1374: How does this avoid a copy?  `create_send_buffer()` still calls `fc::raw::pack()` with the datastream and the message, which eventually calls datastream's `write()` method which calls `memcpy()`.

jgiszczak (approved):

spoonincode (changes requested):
hold up

plugins/net_plugin/net_plugin.cpp
```diff
@@ -861,18 +894,21 @@ namespace eosio {
...
       auto rep = socket->remote_endpoint(ec);
       auto lep = socket->local_endpoint(ec2);
-      std::lock_guard<std::mutex> g_conn( conn_mtx );
-      remote_endpoint_ip = ec ? unknown : rep.address().to_string();
-      remote_endpoint_port = ec ? unknown : std::to_string(rep.port());
+      log_remote_endpoint_ip = ec ? unknown : rep.address().to_string();
+      log_remote_endpoint_port = ec ? unknown : std::to_string(rep.port());
       local_endpoint_ip = ec2 ? unknown : lep.address().to_string();
       local_endpoint_port = ec2 ? unknown : std::to_string(lep.port());
+      std::lock_guard<std::mutex> g_conn( conn_mtx );
```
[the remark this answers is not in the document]
heifner: `start_listen_loop` accesses all other other connection strands which can be in different threads. It is not it's `remote_endpoint_ip` it is accessing it is other connections.

plugins/net_plugin/net_plugin.cpp
heifner: It is not verifying that uint32_t is 4 bytes so much as it is verifying that the `header_size` remains the same as `message_header_size` since it is used during the read. The idea is if you change this code you better also change all uses of `message_header_size`.

plugins/net_plugin/net_plugin.cpp
heifner: See comment below

plugins/net_plugin/net_plugin.cpp
heifner: `net_message` is a `variant< .., signed_block, ..>`. To call `create_send_buffer( const net_message& m )` you would need to construct a `net_message` which would require copying the `signed_block` into the `variant`. 
Here we call `create_send_buffer( uint32_t which, const T& v )` which creates the `vector<char>` from the `which` and the `signed_block` without creating a `net_message`.

plugins/net_plugin/net_plugin.cpp
jgiszczak: payload_size is a uint32_t.  By definition (in the standard), uint32_t is a 4 byte integer.  So `sizeof(payload_size)` is 4, always.  constexpr size_t header_size == 4.  If the intent is to be sure message_header_size is always 4 bytes, that's a ridiculously convoluted way to achieve it.  Better to just change the definition: `constexpr auto message_header_size = sizeof(uint32_t);`, dispense with `header_size` here and just use `message_header_size` directly to write the bytes of `payload_size`.  That would be more consistent with how it's used during reading anyway.

Meanwhile `static_assert(
    sizeof(decltype(fc::raw::pack_size(std::declval<net_message&>())))==sizeof(uint32_t),
     "pack_size return value doesn't always fit in 4 bytes");` will fail.  (But we're not trying to solve that problem.  I just couldn't let go of the expression after I figured out the syntax...)

spoonincode (approved):
Been playing around with different combinations of nodeos versions connected to each other. Never saw any issues.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1198,21 +1260,22 @@ namespace eosio {
...
          return false;
       } else {
-         fc_dlog( logger, "enqueue sync block ${num}", ("num", peer_requested->last + 1) );
+         peer_dlog( this, "enqueue sync block ${num}", ("num", peer_requested->last + 1) );
       }
       uint32_t num = ++peer_requested->last;
       if(num == peer_requested->end_block) {
          peer_requested.reset();
-         fc_ilog( logger, "completing enqueue_sync_block ${num} to ${p}", ("num", num)("p", peer_name()) );
+         peer_ilog( this, "completing enqueue_sync_block ${num}", ("num", num) );
```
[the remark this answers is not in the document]
heifner: I'll change that to `peer_dlog`, does seem to be a bit chatty.

plugins/net_plugin/net_plugin.cpp
heifner: Updated to use `message_header_size`

PR #173, opened 2022-05-02
Implement return failure traces

jgiszczak:
Original APIs hardcoded false, new compute_transaction hardcoded true, new send_transaction2 now honors parameter.

Updates Python test framework to interpret returned failure traces.


tests/nodeos_run_test.py
```diff
@@ -549,7 +549,7 @@
     action="transfer"
     data="{\"from\":\"defproducera\",\"to\":\"currency1111\",\"quantity\":"
     data +="\"00.0025 CUR\",\"memo\":\"test\"}"
-    opts="--permission defproducera@active"
+    opts="--permission defproducera@active --use-old-send-rpc"
```
heifner, line 552: Why do these need to use the old rpc? Can that be updated?

jgiszczak: They could now be updated but I thought it would be useful to include at least a couple of tests that still use the other RPC API, so it doesn't stop being tested.  As long as it exists, we should continue verifying its functionality.

heifner: Can you add a comment to that affect, so it is obvious this is here to test that and can be easily removed in the future.

jgiszczak: Comment added.

heifner (approved):

PR #174, opened 2022-05-02
small tweak to eos-vm to build with gcc 12

spoonincode:
Just a small change from over here: eosnetworkfoundation/mandel-eos-vm#6


swatanabe (approved):

spoonincode:
eosnetworkfoundation/mandel-eos-vm#6 was merged before eosnetworkfoundation/mandel-eos-vm#4 and I was careful to only move the submod in this branch to the first merged commit so as not to interfere with #184 which will clearly go in after this one since it's not approved yet.

PR #185, opened 2022-05-04
fixing flakiness in voting and forked chain tests

ClaytonCalabrese:
### Resolve flakiness in nodeos_voting_lr_test and nodeos_forked_chain_lr_test
This is a port of a fix from the EOS PR 11022: https://github.com/EOSIO/eos/pull/11022

Fixes:

1. nodeos_voting_lr_test. It was broken after bugfix in `Node.waitForIrreversibleBlock` where it was waiting for head instead of lib. After fix this increased execution time and lead to timeout fails.
2. nodeos_forked_chain_lr_test. It had flakiness due to race condition in head catchup and switching forks that lead to `go_away_message` being sent with status `benign_other` due to unknown block requested. Fix adding retry after closing connection.

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -2755,7 +2755,11 @@ namespace eosio {
       }
       if( msg.reason == wrong_version ) {
          if( !retry ) no_retry = fatal_other; // only retry once on wrong version
-      } else {
+      }
+      else if ( msg.reason == benign_other ) {
+         if ( retry ) fc_dlog( logger, "received benign_other reason, retrying to connect");
+      }
+      else {
          retry = false;
       }
       flush_queues();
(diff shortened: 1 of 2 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -2755,7 +2755,11 @@ namespace eosio {
       }
       if( msg.reason == wrong_version ) {
          if( !retry ) no_retry = fatal_other; // only retry once on wrong version
-      } else {
+      }
+      else if ( msg.reason == benign_other ) {
+         if ( retry ) fc_dlog( logger, "received benign_other reason, retrying to connect");
```
heifner, line 2760: Change this to a `peer_dlog( this, ...`

heifner (changes requested):

heifner (approved):

Issue #191, opened 2022-05-06
port eos pr #11017 to add decompressed_size_under_limit test case


ClaytonCalabrese:
This would add a test case to unittests/chain_tests.cpp to ensure that zlib isn't causing bloat on decompression of transactions.

PR #192, opened 2022-05-06
backport decompressed_size_under_limit test case

ClaytonCalabrese:
### This PR is a backport of https://github.com/EOSIO/eos/pull/11017

In doing so, portions of the following PRs were added:
https://github.com/EOSIO/eos/pull/8988
https://github.com/EOSIO/eos/pull/9204
https://github.com/EOSIO/eos/pull/9447

These PRs are what construct `test_cfd_transaction.hpp` and `test_cfd_transaction.cpp`. No other portion of these PRs aside from these files and the parts from `transaction.cpp`, `transaction.hpp`, and `api_tests.cpp` which become part of test_cfd_transaction have been ported.

Resolves issue https://github.com/eosnetworkfoundation/mandel/issues/191

### DIFF
--- a/libraries/chain/include/eosio/chain/transaction.hpp
+++ b/libraries/chain/include/eosio/chain/transaction.hpp
@@ -159,8 +159,6 @@ namespace eosio { namespace chain {
          local_pack_context_free_data();
       }
 
-      packed_transaction(const bytes& packed_txn, const vector<signature_type>& sigs, const bytes& packed_cfd, compression_type _compression);
-
       // used by abi_serializer
       packed_transaction( bytes&& packed_txn, vector<signature_type>&& sigs, bytes&& packed_cfd, compression_type _compression );
       packed_transaction( bytes&& packed_txn, vector<signature_type>&& sigs, vector<bytes>&& cfd, compression_type _compression );
--- a/libraries/chain/include/eosio/chain/transaction.hpp
+++ b/libraries/chain/include/eosio/chain/transaction.hpp
@@ -159,6 +159,8 @@ namespace eosio { namespace chain {
          local_pack_context_free_data();
       }
 
+      packed_transaction(const bytes& packed_txn, const vector<signature_type>& sigs, const bytes& packed_cfd, compression_type _compression);
+
       // used by abi_serializer
       packed_transaction( bytes&& packed_txn, vector<signature_type>&& sigs, bytes&& packed_cfd, compression_type _compression );
       packed_transaction( bytes&& packed_txn, vector<signature_type>&& sigs, vector<bytes>&& cfd, compression_type _compression );
--- a/unittests/chain_tests.cpp
+++ b/unittests/chain_tests.cpp
@@ -104,16 +104,11 @@ BOOST_AUTO_TEST_CASE( decompressed_size_over_limit ) try {
    bytes pcfd = pt.get_packed_context_free_data();
    vector<signature_type>  sigs;
    sigs.push_back(sig);
-   bool throw_expect_string = false;
-   try {
-      // will call into decompress zlib
-      packed_transaction copy( std::move(packed_txn), std::move(sigs), std::move(pcfd), packed_transaction::compression_type::zlib );
-   } catch(fc::exception& er) {
-     if ( er.to_detail_string().find("Exceeded maximum decompressed transaction size") != std::string::npos){
-       throw_expect_string = true;
-     }
-   }
-   BOOST_REQUIRE(throw_expect_string);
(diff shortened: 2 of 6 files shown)


libraries/chain/include/eosio/chain/transaction.hpp
```diff
@@ -159,6 +159,8 @@ namespace eosio { namespace chain {
          local_pack_context_free_data();
       }
 
+      packed_transaction(const bytes& packed_txn, const vector<signature_type>& sigs, const bytes& packed_cfd, compression_type _compression);
```
heifner, line 162: Don't add this just for a test. Use the constructor below. 
`packed_transaction( bytes&& packed_txn, vector<signature_type>&& sigs, bytes&& packed_cfd, compression_type _compression );`

unittests/chain_tests.cpp
```diff
@@ -74,4 +76,47 @@ BOOST_AUTO_TEST_CASE( replace_account_keys ) try {
...
+   // add a normal action along with cfa
+   dummy_action         da = {DUMMY_ACTION_DEFAULT_A, DUMMY_ACTION_DEFAULT_B, DUMMY_ACTION_DEFAULT_C};
+   eosio::chain::action act1(
+       std::vector<eosio::chain::permission_level>{{"testapi"_n, eosio::chain::config::active_name}}, da);
+   trx.actions.push_back(act1);
+   chain.set_transaction_headers(trx);
+   auto sig = trx.sign(chain.get_private_key("testapi"_n, "active"), chain.control->get_chain_id());
+
+   // pack
+   packed_transaction ptv0(trx, packed_transaction::compression_type::zlib);
```
heifner, line 101: `ptv0` should be just `pt`. There is not `v0` vs `v1` in Mandle.

heifner (changes requested):

unittests/chain_tests.cpp
```diff
@@ -74,4 +76,47 @@ BOOST_AUTO_TEST_CASE( replace_account_keys ) try {
...
+      // will call into decompress zlib
+      packed_transaction copy( packed_txn, sigs, pcfd, packed_transaction::compression_type::zlib );
+   } catch(fc::exception& er){
+     threw = true;
+     if ( er.to_detail_string().find("Exceeded maximum decompressed transaction size") != std::string::npos){
+       throw_expect_string = true;
+     }
+   }
+   BOOST_REQUIRE(threw == false);
+   BOOST_REQUIRE(throw_expect_string == false);
```
spoonincode, line 119: If we're trying to make sure it doesn't throw, which is how it is written, why do the try/catch at all? Unhandled exceptions will cause the test case to fail.

ClaytonCalabrese:
Looks like I also lost a test case at some point. I'll add it back.

unittests/chain_tests.cpp
ClaytonCalabrese: I assume the try catch exists so we can search the exception for the failure we expect. I removed the variable and BOOST_REQUIRE to not throw at all.

spoonincode: It's still not clear what the intent here is, and maybe @heifner can chime in since it seems like it's from his original PR?

The original code looked like it was trying to check that _nothing_ was thrown but it was doing so in a very roundabout way. This code looks like it allows _anything to be thrown that isn't "Exceeded maximum..."_ which would may or may not change the semantics.

I tried running the test locally and nothing was being thrown.

heifner: I only side-ported it to develop-boxed. It appears it was written this odd way as an inverse of the test above it. My guess is the idea was that it is more obvious it is the opposite of the above test by just changing the BOOST_REQUIRE.

spoonincode: I missed that, it explains it. But yeah that's not so pretty either. Should be something more on the order of
```c++
BOOST_REQUIRE_THROW(..., tx_decompression_error);
```
bonus points for
```c++
BOOST_REQUIRE_THROW(..., tx_decompression_error, [](const tx_decompression_error& e) {
      return e.to_detail_string().find("Exceeded maximum decompressed transaction size") != std::string::npos;
   }));
```

But at least its intent was not confusing like this one is.

spoonincode: If we're going to touch the test (as we've started to), might as well clean it up further. If not, we can just PR changes to it later.

ClaytonCalabrese: I don't think `BOOST_REQUIRE_THROW` has a version which takes 3 arguments.

spoonincode: yep that's the one

heifner (approved):

Issue #207, opened 2022-05-10
nodeos needs more time on linux before it starts accepting p2p transactions


ClaytonCalabrese:

heifner:
Resolved by #309

PR #210, opened 2022-05-10
bios boot tutorial sleep was too short on linux

ClaytonCalabrese:
### port of https://github.com/EOSIO/eos/pull/9738

> The bios-boot-tutorial.py needs to give nodeos additional time starting up before it continues on with execution. If the script moves too fast, nodeos may not be at a point it is accepting p2p transactions and the follow on steps in the script will fail. 10 seconds seems long enough to allow the script to consistently succeed.

issue https://github.com/eosnetworkfoundation/mandel/issues/207


oschwaldp-oci (approved):

oschwaldp-oci (commented):
Port of change looks good. Holding off on your analysis of the tutorial itself.

ClaytonCalabrese:
There are other problems in the tutorial. Futher investigation is needed.

PR #234, opened 2022-05-14
Rewrite ship_tests's ship_client to C++; remove node & npm as dependencies for a test

spoonincode:
Rewrite ship_client used by ship_test.py to C++. This removes the need to have node and npm installed and configured properly to run the full test suite.

I’m not against having tests which are optional based on whether dependencies are installed or not, but what made this one particularly insidious is that it didn't just require node to be installed: it also required npm to be installed and for the user to know some npm incantation to properly install the packages so the test will run. If you just happen to have node installed, which clutched in the test, it can be hard for a node/npm novice to know how to get the test to start passing.

There is also some ancillary fallout, like some automated tools picking up on the package.json in the repo's root and starting to judge if our npm dependencies are up-to-date and such. (It doesn't matter)

After getting in to this rewrite I realized ship_test (and by extension ship_client), well, it just doesn't do much. ship_client just bangs on the get_status_request_v0 request. ship_client originally printed out some timing diagnostics but those aren't consumed by ship_test so I didn't bother porting them. The important aspect of the test I wanted to retain -- what I believe is a significant part of its spirit -- was that it consumes the ABI from the websocket and uses it (i.e. it doesn't use ship_protocol.hpp). I'm leveraging abieos to perform that task, and I've stuck abieos' submodule down in to the tests directory to try and make it clear its use is isolated at the moment.


tests/ship_client.cpp
```diff
@@ -0,0 +1,143 @@
...
+   bpo::notify(varmap);
+
+   if(help) {
+      cli.print(std::cout);
+      return 0;
+   }
+
+   std::string::size_type colon = socket_address.find(':');
+   std::string statehistory_server = socket_address.substr(0, colon);
+   std::string statehistory_port = socket_address.substr(colon+1);
```
heifner, line 44: Possible `string::npos + 1` here if no `:` in `-a` which will be `0` and return an empty string, but seems a better error could be provided if `colon` is checked for `npos`.

heifner (changes requested):

heifner (approved):

PR #247, opened 2022-05-15
look for openssl 3.0 & 1.1 by default on macos

spoonincode:
If I `brew install openssl` on a fresh Mac I get OpenSSL 3.0. Look for either 3.0 or 1.1 by default on macOS


heifner (approved):

spoonincode:
hm, we maybe should just simply use `openssl`... but this way we always prefer openssl 3.0, so maybe that's good too

Issue #248, opened 2022-05-15
test_trx_finality_status_processing test failing on Mac


heifner:
```
mandel/plugins/chain_plugin/test/test_trx_finality_status_processing.cpp:196: error: in "trx_finality_status_processing_test/trx_finality_status_logic": check cs.head_id == *std::get<0>(trx_pairs_20[0])->producer_block_id has failed
info  2022-04-04T04:44:11.500 thread-0  trx_finality_status_pr:235    ensure_storage       ] Finality Status exceeded max storage (0GB) need to free up 0 GB
info  2022-04-04T04:44:11.500 thread-0  trx_finality_status_pr:297    ensure_storage       ] Finality Status dropped 20 transactions, which were removed from block # 1 to block # 2
```

PR #250, opened 2022-05-15
Fix test_trx_finality_status_processing test

heifner:
The test was dereferencing an empty `std::optional`. Fixed the code to verify that the optional has no value.


spoonincode (approved):
Fixed failure I was seeing on macos

spoonincode:
fixes #248

PR #251, opened 2022-05-15
escape ']' character in deep mind unit test regexes

spoonincode:
The `]` character was not escaped in the regexes that the deep mind test uses to create `std::regex` instances. This was leading to a
```
std::__1::regex_error: The parser did not consume the entire regular expression.
```
error on libc++. libstdc++ was seemingly okay though.

I won't claim to be an expert in ECMAScript regex grammar to say which one of the stdlibs is "correct". But escaping both seems proper, and it makes both happy.


swatanabe (approved):
Leaving it out was a typo.

Issue #243, opened 2022-05-14
Default to returning WASM as returning WAST has been deprecated.


heifner:
As there is no more support for `get_code` to return WAST the default to return WASM should be `true`.

Backport https://github.com/EOSIO/eos/pull/9706

PR #272, opened 2022-05-18
Default to returning WASM as returning WAST has been deprecated.

oschwaldp-oci:
Default to returning WASM as returning WAST has been deprecated.

As there is no more support for get_code to return WAST the default to return WASM should be true.

Backport https://github.com/EOSIO/eos/pull/9706

Resolves https://github.com/eosnetworkfoundation/mandel/issues/243

### DIFF
--- a/plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
+++ b/plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
@@ -216,7 +216,7 @@ public:
 
    struct get_code_params {
       name account_name;
-      bool code_as_wasm = false;
+      bool code_as_wasm = true;
    };
 
    struct get_code_hash_results {


heifner (approved):

spoonincode (changes requested):
one sec

heifner:
> one sec

It asserts if you pass `false`, so might as well default to `true`. There is a diff fix for removing wast support completely.

spoonincode:
yeah this change doesn't really do anything material. There needs to be an issue created for EOSIO/eos#10819 still.

heifner:
> yeah this change doesn't really do anything material. There needs to be an issue created for [EOSIO/eos#10819](https://github.com/EOSIO/eos/pull/10819) still.

Exactly.

spoonincode:
I guess it does allow an HTTP request made (w/o cleos) to lack that field

spoonincode (approved):

Issue #242, opened 2022-05-14
Add integration test for nodeos shutdown 


heifner:
Add an integration of nodeos for crash when the nodes are killed.

https://github.com/EOSIO/eos/pull/9494

Description:
Added a new integration test for nodeos. This test sets up one producing nodes and one "bridge" node using test_control_api_plugin. We send transactions to the bridge node. After the accounts are initialized and allocated with symbols, we kill the producing node. Then we flood hundreds transactions to the bridge node. At the end, we ill the bridge. It tests that no crashes happen when the nodes are killed.

PR #281, opened 2022-05-19
Add integration test for nodeos shutdown

oschwaldp-oci:
Add an integration of nodeos for crash when the nodes are killed.

Backports https://github.com/EOSIO/eos/pull/9494

Description:
Added a new integration test for nodeos. This test sets up one producing nodes and one "bridge" node using test_control_api_plugin. We send transactions to the bridge node. After the accounts are initialized and allocated with symbols, we kill the producing node. Then we flood hundreds transactions to the bridge node. At the end, we ill the bridge. It tests that no crashes happen when the nodes are killed.

Resolves https://github.com/eosnetworkfoundation/mandel/issues/242

### DIFF
--- a/tests/nodeos_contrl_c_test.py
+++ b/tests/nodeos_contrl_c_test.py
@@ -1,0 +1,106 @@
+#!/usr/bin/env python3
+
+from testUtils import Utils
+from datetime import datetime
+from datetime import timedelta
+import time
+from Cluster import Cluster
+import json
+from WalletMgr import WalletMgr
+from Node import Node
+from TestHelper import TestHelper
+
+import signal
+
+###############################################################
+# nodeos_contrl_c_lr_test
+#
+# This test sets up one producing nodes and one "bridge" node using test_control_api_plugin. We send
+# transactions to the bridge node. After the accounts are initialized and allocated with symbols, we
+# kill the producing node. Then we flood hundreds transactions to the bridge node. At the end, we
+# kill the bridge. It tests that no crashes happen when the nodes are killed.
+#
+###############################################################
+
+errorExit=Utils.errorExit
+
+args = TestHelper.parse_args({"--wallet-port", "-v"})
+
+cluster=Cluster(walletd=True)
+killAll=True
+totalProducerNodes=1
+totalNonProducerNodes=1
+totalNodes=totalProducerNodes+totalNonProducerNodes
+walletPort=args.wallet_port
+walletMgr=WalletMgr(True, port=walletPort)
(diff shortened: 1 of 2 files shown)


tests/nodeos_contrl_c_test.py
```diff
@@ -48,15 +50,17 @@
     # producer nodes will be mapped to 0 through totalProducerNodes-1, so the number totalProducerNodes will be the non-producing node
     specificExtraNodeosArgs[totalProducerNodes] = "--plugin eosio::producer_plugin --plugin eosio::chain_api_plugin --plugin eosio::http_plugin "
     "--plugin eosio::txn_test_gen_plugin --plugin eosio::producer_api_plugin "
+    traceNodeosArgs = " --plugin eosio::trace_api_plugin --trace-no-abis "
```
heifner: Does Cluster require this now to work at all. If so, we should create an issue to just have Cluster automatically add this if not already added.

tests/nodeos_contrl_c_test.py
```diff
@@ -0,0 +1,115 @@
...
+        trans=nonProdNode.delegatebw(account, 20000000.0000, 20000000.0000, waitForTransBlock=True, exitOnError=True)
+
+    testSuccessful = prodNode.kill(signal.SIGTERM)
+    if not testSuccessful:
+        TestHelper.shutdown(cluster, walletMgr, testSuccessful=testSuccessful, killEosInstances=True, killWallet=True, keepLogs=True, cleanRun=True, dumpErrorDetails=True)
+        errorExit("Failed to kill the producer node")
+
+    for amt in range(1, 500, 1):
+        xferAmount = Node.currencyIntToStr(amt, CORE_SYMBOL)
+        nonProdNode.transferFunds(accounts[0], accounts[1], xferAmount, "test transfer", waitForTransBlock=False)
```
heifner, line 103: Use the new `transferFundsAsync` like in `nodeos_retry_transaction_test.py` which should hopefully get a few more in flight at time of shutdown.

heifner (changes requested):

tests/nodeos_contrl_c_test.py
oschwaldp-oci: See initial notes in new issue: https://github.com/eosnetworkfoundation/mandel/issues/285

tests/nodeos_contrl_c_test.py
oschwaldp-oci: Done.

heifner (approved):

Issue #303, opened 2022-05-23
Feature Hashes Do Not Match Between eosio and Mandel


ClaytonCalabrese:
on eosio the hash for BLOCKCHAIN_PARAMETERS is 5443fcf88330c586bc0e5f3dee10e7f63c76c00249c87fe4fbf7f38c082006b4 
on mandel the hash for BLOCKCHAIN_PARAMETERS is 8e466518f3e16a679ffa80d2e810e4097e89b0495ed70c112375a1f525093f33 

The hash is different due to the difference in an enum.

In mandel that is:

enum class builtin_protocol_feature_t : uint32_t {
   preactivate_feature,
   only_link_to_existing_permission,
   replace_deferred,
   no_duplicate_deferred_id,
   fix_linkauth_restriction,
   disallow_empty_producer_schedule,
   restrict_action_to_self,
   only_bill_first_authorizer,
   forward_setcode,
   get_sender,
   ram_restrictions,
   webauthn_key,
   wtmsig_block_signatures,
   action_return_value,
   configurable_wasm_limits,
   blockchain_parameters,
   get_code_hash,
};

In eosio/eos it is:

enum class builtin_protocol_feature_t : uint32_t {
   preactivate_feature,
   only_link_to_existing_permission,
   replace_deferred,
   no_duplicate_deferred_id,
   fix_linkauth_restriction,
   disallow_empty_producer_schedule,
   restrict_action_to_self,
   only_bill_first_authorizer,
   forward_setcode,
   get_sender,
   ram_restrictions,
   webauthn_key,
   wtmsig_block_signatures,
   action_return_value,
   kv_database,
   configurable_wasm_limits,
   blockchain_parameters,
   security_group,
   resource_payer,
   get_wasm_parameters_packed_fix
};

Note blockchain_parameters 15 in Mandel and 16 in eosio/eos, hence different hash for features defined after kv_database, because it is not in Mandel.  

As a result, this enum should use literal numbers to define features with the same hash.

PR #308, opened 2022-05-23
Explicit Values for builtin_protocol_feature_t enum

ClaytonCalabrese:
### Use Explicit Values for builtin_protocol_feature_t 
Feature hashes were not consistent with pre mandel ones, because the enum `builtin_protocol_feature_t ` does not have `kv_database`, which if it did would increase the value of `configurable_wasm_limits` and `blockchain_parameters` by 1.  `configurable_wasm_limits` will be different anyways because mandel uses configurable_wasm_limits2, but blockchain paramenters should have been the same hash on the fork.

Resolves: #303


libraries/chain/include/eosio/chain/protocol_feature_manager.hpp
```diff
@@ -12,23 +12,25 @@ enum class protocol_feature_t : uint32_t {
 };
```
heifner, line 13: Add comment to both `builtin_protocol_feature_t` and `protocol_feature_t` that these values are included in the protocol digest so there values must remain constant.

libraries/chain/include/eosio/chain/protocol_feature_manager.hpp
```diff
@@ -12,23 +12,25 @@ enum class protocol_feature_t : uint32_t {
...
+   only_bill_first_authorizer = 7,
+   forward_setcode = 8,
+   get_sender = 9,
+   ram_restrictions = 10,
+   webauthn_key = 11,
+   wtmsig_block_signatures = 12,
+   action_return_value = 13,
+   // 14 reserved
+   configurable_wasm_limits = 15,
+   blockchain_parameters = 16,
```
heifner, line 31: Add comment to this that it is `16` to match 2.1 value.

heifner (changes requested):

heifner:
> If jungle has already activated these features, this change will force a reset.

Jungle has not activated these features, I asked.

libraries/chain/include/eosio/chain/protocol_feature_manager.hpp
```diff
@@ -12,23 +12,25 @@ enum class protocol_feature_t : uint32_t {
...
+   restrict_action_to_self = 6,
+   only_bill_first_authorizer = 7,
+   forward_setcode = 8,
+   get_sender = 9,
+   ram_restrictions = 10,
+   webauthn_key = 11,
+   wtmsig_block_signatures = 12,
+   action_return_value = 13,
+   // 14 reserved
+   configurable_wasm_limits = 15,
```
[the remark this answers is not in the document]
heifner: There is changes to the description which provides a different feature digest already for `configurable_wasm_limits`, but yeah likely better to make it diff here as well.

heifner (approved):

PR #317, opened 2022-05-26
ARM64 Builds and Pinned Builds

larryk85:
I consolidated both the ARM64 building and pinned builds.

This allows for building of Mandel on non-AMD64 machines and a couple of utility scripts to create a pinned build for creating stable x86_64 binary packages and a script for creating the ARM64 packages for those machines.


larryk85:
@spoonincode went you get to merging your packaging PR, just remove the cpack stuff at the bottom of CMakeLists.txt.

scripts/arm_build.sh
```diff
@@ -0,0 +1,84 @@
...
+
+install_boost
+
+
+# build Mandel
+
+echo "Building Mandel"
+pushdir ${MANDEL_DIR}
+
+try cmake -S${SCRIPT_DIR}/.. -DCMAKE_BUILD_TYPE=Release -DENABLE_OC=Off ..
```
spoonincode, line 65: Is this `ENABLE_OC=Off` needed? OC should have already been disabled?
https://github.com/eosnetworkfoundation/mandel/blob/27e15012af77a2aea3cd02a1a21c5404ed32a215/CMakeLists.txt#L72-L73

More generally, is this entire script needed? What does it offer the user over just cmake && make?

larryk85: It is not necessarily needed.  Neither script is really needed we can just document what they need to do to produce the packages.

This just allows us to write up for the group who are going to produce the packages for this release.

run install_deps.h then pinned_build.sh and on ARM run install_deps.h and arm_build.sh with roughly the same inputs.

When we support this better or envelope this as part of CICD then we can remove all of these and simply the steps in the CICD recipe or Dockerfile.

scripts/pinned_build.sh
```diff
@@ -0,0 +1,136 @@
...
+fi
+
+CORE_SYM=EOS
+DEP_DIR=$1
+MANDEL_DIR=$2
+JOBS=$3
+CLANG_VER=11.0.1
+BOOST_VER=1.70.0
+LLVM_VER=7.1.0
+LIBPQXX_VER=7.2.1
```
spoonincode, line 25: no need for libpqxx variable

scripts/pinned_build.sh
```diff
@@ -0,0 +1,136 @@
...
+      exit -1
+fi
+
+CORE_SYM=EOS
+DEP_DIR=$1
+MANDEL_DIR=$2
+JOBS=$3
+CLANG_VER=11.0.1
+BOOST_VER=1.70.0
+LLVM_VER=7.1.0
```
spoonincode, line 24: How did these versions come about? Some of them are older than what was used in the 2.0.x pinned builds even.

larryk85: Feel free to recommend versions. I honestly didn't really look at the B1 stuff as personally I would've liked the pinned packages to use as low of a version as we support without running into issues or performance degradation as they should be the most stable at this point.

I know the clang version is a bit high, but the delta was noticeable enough that I wanted that one to be at 11.

scripts/pinned_build.sh
larryk85: Thanks I missed that one.

spoonincode:
the change in abieos doesn't seem to work

```
/home/moon/mandel/tests/abieos/src/key_test.cpp: In function ‘void test_compare()’:
/home/moon/mandel/tests/abieos/src/key_test.cpp:144:76: error: narrowing conversion of ‘'\377'’ from ‘char’ to ‘signed char’ [-Wnarrowing]
  144 |    test_key(std::vector<signed char>{'\0'}, std::vector<signed char>{'\xFF'});
      |                                                                            ^
/home/moon/mandel/tests/abieos/src/key_test.cpp:145:76: error: narrowing conversion of ‘'\377'’ from ‘char’ to ‘signed char’ [-Wnarrowing]
  145 |    test_key(std::vector<signed char>{'\1'}, std::vector<signed char>{'\xFF'});
      |
```

I would be willing to make a change to stop rapidjson using Werror if you'd like. Don't like that _at all_. Didn't realize it was going that :-1:

larryk85:
> the change in abieos doesn't seem to work
> 
> ```
> /home/moon/mandel/tests/abieos/src/key_test.cpp: In function ‘void test_compare()’:
> /home/moon/mandel/tests/abieos/src/key_test.cpp:144:76: error: narrowing conversion of ‘'\377'’ from ‘char’ to ‘signed char’ [-Wnarrowing]
>   144 |    test_key(std::vector<signed char>{'\0'}, std::vector<signed char>{'\xFF'});
>       |                                                                            ^
> /home/moon/mandel/tests/abieos/src/key_test.cpp:145:76: error: narrowing conversion of ‘'\377'’ from ‘char’ to ‘signed char’ [-Wnarrowing]
>   145 |    test_key(std::vector<signed char>{'\1'}, std::vector<signed char>{'\xFF'});
>       |
> ```
> 
> I would be willing to make a change to stop rapidjson using Werror if you'd like. Don't like that _at all_. Didn't realize it was going that 👎

Yeah that works, I'm good with that.

spoonincode:
ah sorry it is actually _not_ Werror

> Specifically, G++ gives an error for narrowing conversions from constants, but by default only gives a warning for non-constants.

larryk85:
> ah sorry it is actually _not_ Werror
> 
> > Specifically, G++ gives an error for narrowing conversions from constants, but by default only gives a warning for non-constants.

Gotcha, okay I'll add a change for that in a bit to make it not conliterally or a literal.

spoonincode:
The default wasm runtime isn't being set correctly for an ARM build. Just starting nodeos:

```
warn  2022-05-27T19:09:34.698 nodeos    chain_plugin.cpp:1220         plugin_initialize    ] 3070000 wasm_exception: WASM Exception
eos_vm_jit wasm runtime not supported on this platform and/or configuration
    {"r":"eos_vm_jit"}
    nodeos  wasm_interface_private.hpp:79 wasm_interface_impl

Failed to initialize
```

`test_abieos_template` is also failing on an ARM build
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  Expected number or boolean

```

libraries/chain/include/eosio/chain/config.hpp
```diff
@@ -114,7 +114,12 @@ const static uint32_t   setcode_ram_bytes_multiplier       = 10;     ///< multip
 
 const static uint32_t   hashing_checktime_block_size       = 10*1024;  /// call checktime from hashing intrinsic once per this number of bytes
 
+#ifdef __x86_64__
```
spoonincode, line 117: This might be better checking for `EOSIO_EOS_VM_JIT_RUNTIME_ENABLED`

larryk85: Ah, definitely. I had forgotten we had those. Thanks for that.

scripts/arm_build.sh
spoonincode: It's named arm_build.sh and literally does nothing specific for ARM afaict. It's an 80+ line script to perform the 3 commands already documented in the readme (`apt-get install`, `cmake`, `make`).

It's seemingly exactly the kind of hollow scripts we were trying to move away from.

scripts/pinned_build.sh
spoonincode: It's a bit unexpected because the versions of LLVM & boost chosen are over three years old, older than we used in 2.0.x, and older than what ships in Ubuntu 20.04 (our only supported version for 3.1). But yes they are compatible so there isn't anything really 'wrong' with it, just curious what the criteria was for picking them.

FWIW there are parts of the code that are known to perform notably better with newer version of boost, for example EOSIO/eos#8147 (which uhhh why is that not in mandel)

spoonincode (approved):

PR #312, opened 2022-05-24
Backport of unpacking packed transaction in cleos

ndcgundlach:
Makes cleos unpack a packed txn before signing.

Resolves: https://github.com/eosnetworkfoundation/mandel/issues/291

Original PR: https://github.com/EOSIO/eos/pull/10203

PR #323, opened 2022-05-28
add more --no-auto-keosd flags to cli_test

spoonincode:
cli_test was causing cleos to crash for me (and failing the test). Something to do with `ensure_keosd_running()`. Wait... why is it doing that.. I thought I stopped that in #97. Looks like #312 pulled in more tests here that need the flag. So add the flag.

I probably should investigate why cleos was _crashing_ some time.

### DIFF
--- a/tests/cli_test.py
+++ b/tests/cli_test.py
@@ -87,7 +87,7 @@ def cleos_sign_test():
         '"context_free_data": []'
     '}')
 
-    output = subprocess.check_output(['./programs/cleos/cleos', 'sign',
+    output = subprocess.check_output(['./programs/cleos/cleos', '--no-auto-keosd', 'sign',
                                       '-c', chain, '-k', key, trx])
     # make sure it is signed
     assert(b'signatures' in output)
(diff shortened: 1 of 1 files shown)


heifner (approved):

spoonincode:
> I probably should investigate why cleos was _crashing_ some time.

It's always crashing at
https://github.com/eosnetworkfoundation/mandel/blob/27e15012af77a2aea3cd02a1a21c5404ed32a215/programs/cleos/main.cpp#L1049

even with something as simple as `wallet list`
```
./bin/cleos wallet list
/usr/src/debug/gcc-build/x86_64-pc-linux-gnu/libstdc++-v3/include/bits/basic_string.h:1280: std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::reference std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::front() [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>; reference = char&]: Assertion '!empty()' failed.
```

I'm somewhat suspecting this is some gcc12 + boost 1.78 incompatibility at the moment, so not adding a ticket for it yet

spoonincode:
fwiw upgrading to boost 1.79 seemed to make the problem go away.

PR #324, opened 2022-05-28
don't run appbase's signal catching thread detached after main() exits

spoonincode:
What appbase was newly doing to work around EINTR -- running a detached thread after main() exits -- I'm not sure is defined behavior. Empirically, IME, it can lead to wonky problems even on Linux. And indeed it was leading to a wonky problem on macOS where nodeos wouldn't exit.

It is trivial to refactor this code to not rely on that possibly UB and just stop the signal catching thread on shutdown.

There is some ongoing discussion on how to refactor parts of our code to not need this workaround in the first place. But those changes won't be ready for 3.1 and this, I believe, settles the last known problem with 3.1 & macOS.

Needs eosnetworkfoundation/mandel-appbase#4

Also this is consuming submodule changes from #277 (approved in  eosnetworkfoundation/mandel-appbase#3) since that has been lingering for some time.


swatanabe:
It isn't undefined behavior.  Running a detached thread after main exits is allowed with significant caveats.  Specifically, any call to a non-async-signal-safe function must happen before the end of main.  POSIX requires `pause` to be async-signal-safe.  With that being said, it doesn't really surprise me that there are problems.

heifner (approved):

PR #326, opened 2022-05-30
mark the deb & rpm packages as conflicting with 'eosio' package

spoonincode:
Packages are currently being named "mandel"; these won't upgrade old pre-ENF "eosio" packages, so mark them as a conflict to prevent installation.


spoonincode:
Actually CONFLICTS works really well -- on ubuntu it even uninstalls eosio automatically on the upgrade with no nagging

jgiszczak (approved):

PR #327, opened 2022-05-31
ship log pruning: optionally reduce disk usage of ship log files during operation by periodically trimming logs

spoonincode:
Add a new option, `state-history-log-prune-blocks`, which when set will periodically reduce the disk usage of the state history log files to only that of the given number of most recent blocks. This allows operators to limit the disk usage of state history without stopping and restarting nodeos.

Pruning is performed by "poking holes" in the log files; in other words making them sparse files. This may have implications if trimmed log files are tarballed/copied/etc with an application that does not recognize sparse files: the files may take up more disk space than expected on their destination if an application is not spare file aware. Logs can be "vacuumed" (converted back in to an unpruned log that doesn't contain holes) by emitting the `state-history-log-prune-blocks` option.

The pruning is only periodically performed so in practice a small number of additional blocks (no more than a few megabytes worth of data) will be available beyond the configured number.


libraries/state_history/include/eosio/state_history/log.hpp
```diff
@@ -277,22 +405,91 @@ class state_history_log {
...
          boost::filesystem::resize_file(log_filename, pos);
-         boost::filesystem::resize_file(index_filename, (block_num - _begin_block) * sizeof(uint64_t));
+         boost::filesystem::resize_file(index_filename, (block_num - _index_begin_block) * sizeof(uint64_t));
          _end_block = block_num;
       }
+      log.seek_end(0);
+      ilog("fork or replay: removed ${n} blocks from ${name}.log", ("n", num_removed)("name", name));
+   }
+
+   void vacuum() {
```
arhag, line 423: Add comment for precondition that the log file is in pruned format.

libraries/state_history/include/eosio/state_history/log.hpp
```diff
@@ -277,22 +405,91 @@ class state_history_log {
...
+
+      std::vector<char> buff;
+      buff.resize(4*1024*1024);
+
+      auto tick = std::chrono::time_point_cast<std::chrono::seconds>(std::chrono::system_clock::now());
+      while(copy_sz) {
+         const size_t copy_this_round = std::min(buff.size(), copy_sz);
+         log.seek(copy_from_pos);
+         log.read(buff.data(), copy_this_round);
+         log.punch_hole(copy_from_pos, copy_from_pos+copy_this_round);
```
arhag, line 457: Since `copy_from_pos` is changing in each iteration of the loop, doesn't this mean that the hole punching may not be contiguous if the positions are not precisely aligned with the file block size?

I understand this only matters in terms of temporary disk usage during the vacuum processing since the file is resized at the end anyway. But the temporary disk usage is why you are bothering to do this hole punching here in the first place. And the on-disk data can grow with each iteration of this loop to some fraction of the original disk size unless the positions are precisely aligned with the file block size.

What if the start of the hole punching range was instead `std::max(copy_to_pos + copy_this_round, orig_copy_from_pos)` where `orig_copy_from_pos` is the original `copy_from_pos` value before the start of the while loop? Would this be inefficient for the OS if the range is large but spans sub-ranges that have already been punched out?

libraries/state_history/include/eosio/state_history/log.hpp
```diff
@@ -277,22 +405,91 @@ class state_history_log {
...
+
+         for(uint32_t new_block_num = 0; new_block_num < num_blocks_in_log; ++new_block_num) {
+            const uint64_t new_pos = index_ptr[new_block_num + offset_blocks] - offset_bytes;
+            index_ptr[new_block_num] = new_pos;
+
+            if(new_block_num + 1 != num_blocks_in_log)
+               log.seek(index_ptr[new_block_num + offset_blocks + 1] - offset_bytes - sizeof(uint64_t));
+            else
+               log.seek_end(-sizeof(uint64_t));
+            log.write((char*)&new_pos, sizeof(new_pos));
```
arhag, line 486: Hmm... maybe we should have done relative offset positioning for the entry suffixes. Oh well, not feasible to change now.

arhag (changes requested):

libraries/state_history/include/eosio/state_history/log.hpp
```diff
@@ -110,22 +167,41 @@ class state_history_log {
...
+
+      if(prune_blocks) {
+         //only bother to try pruning every 4MB written. except for when the config is set very low (mainly for unit test purposes)
+         uint64_t bother_every = 4*1024*1024;
+         if(*prune_blocks < 16)
+            bother_every = 4096;
+
+         const uint64_t mask = ~(bother_every-1);
+         if((pos&mask) != (end&mask))
+            prune();
```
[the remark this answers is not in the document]
spoonincode: > `prune` won't be called if each entry is less than 4 MiB no matter how many such entries are written. There is no accumulated tracking across `write_entry` calls.

Why wouldn't it? Let's consider bother_every=256 which makes mask 0xFF, and writing 4 entries that are all smaller than than 256.
```
pos: 0x000 end:0x040   maskedpos: 0x000 maskedend: 0x000
pos: 0x040 end:0x0a0   maskedpos: 0x000 maskedend: 0x000
pos: 0x0a0 end:0x0e0   maskedpos: 0x000 maskedend: 0x000
pos: 0x0e0 end:0x120   maskedpos: 0x000 maskedend: 0x100  maskedpos != maskedend, so prune()
```

libraries/state_history/include/eosio/state_history/log.hpp
spoonincode: hmm.. I need to think it through a little more. Maybe even something as simple as
```c++
log.punch_hole(copy_to_pos, copy_from_pos+copy_this_round);
```
is fine.

libraries/state_history/include/eosio/state_history/log.hpp
```diff
@@ -110,22 +167,41 @@ class state_history_log {
...
+
+      //if we're operating on a pruned block log and this is the first entry in the log, make note of the feature in the header
+      if(prune_blocks && _begin_block == _end_block)
+         header.magic = ship_magic(get_ship_version(header.magic), ship_feature_pruned_log);
+
       uint64_t pos = log.tellp();
       write_header(header);
       write_payload(log);
       uint64_t end = log.tellp();
-      EOS_ASSERT(end == pos + state_history_log_header_serial_size + header.payload_size, chain::plugin_exception,
```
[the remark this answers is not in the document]
spoonincode: Unneeded afaict. `cfile` already checks the return value on all its writes that it wrote everything it expected and throws an exception otherwise.

If you go back in time you'll see that originally `log` was a `std::fstream` and the code looked like
https://github.com/EOSIO/eos/blob/c7f026710038f91f39c2a0d293d7102654c9576f/plugins/state_history_plugin/include/eosio/state_history_plugin/state_history_log.hpp#L101-L106
In this case if the `std::fstream::write()` fails no exception is thrown you'd need to go inspect the stream's state and stuff. So this EOS_ASSERT appears to be there to check that everything was written. When the code was modified from `std::fstream` to `fc::cfile` this could have been removed.

libraries/state_history/include/eosio/state_history/log.hpp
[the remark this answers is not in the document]
spoonincode: Yeah all the pruning options should be bundled in to a config struct. I'll get to that tomrrow.

libraries/state_history/include/eosio/state_history/log.hpp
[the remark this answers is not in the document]
arhag: I think the issue is that there could be a mismatch between the `payload_size` in the `header` and the actual bytes written out by the `write_payload` function. That is the responsibility of the caller to ensure they match, but this check would catch it early if they don't.

libraries/state_history/include/eosio/state_history/log.hpp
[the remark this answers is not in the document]
arhag: Yeah, I see. But it seems there is still a flaw if it is aligned in an unfortunate way. For example:

```
pos: 0x000 end:0x038   maskedpos: 0x000 maskedend: 0x000
pos: 0x040 end:0x078   maskedpos: 0x000 maskedend: 0x000
pos: 0x080 end:0x0b8   maskedpos: 0x000 maskedend: 0x000
pos: 0x0c0 end:0x0f8   maskedpos: 0x000 maskedend: 0x000
pos: 0x100 end:0x138   maskedpos: 0x100 maskedend: 0x100
```

I think that can be resolved by ensuring the end also includes the 8 bytes suffix so that the tracked data ranges always remain contiguous.

libraries/state_history/include/eosio/state_history/log.hpp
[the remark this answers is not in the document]
spoonincode: yep fair enough :+1:

libraries/state_history/include/eosio/state_history/log.hpp
```diff
@@ -59,14 +59,18 @@ struct state_history_log_header {
 static const int state_history_log_header_serial_size = sizeof(state_history_log_header::magic) +
                                                         sizeof(state_history_log_header::block_id) +
                                                         sizeof(state_history_log_header::payload_size);
+struct state_history_log_prune_config {
+   uint32_t                prune_blocks;                  //number of blocks to prune to when doing a prune
+   size_t                  prune_threshold = 4*1024*1024; //(approximately) how many bytes need to be added before a prune is performed
+   std::optional<size_t>   vacuum_on_close;               //when set, a vacuum is performed on dtor if log contains less than this many bytes
```
arhag, line 65: Did you not want to keep the default behavior of vacuuming on exit assuming the remaining size is less than 1 GiB? That behavior has now changed with these latest commits since this config is not set accordingly in state_history_plugin.cpp either (only in the tests).

spoonincode: I should have maintained that just for consistency on the update. But there appears to still be some indecision on what we should do wrt automatic vacuuming. (I thought I was well opinionated on the matter before so many other opinions got mixed in). Let's decide for certain on 13th

arhag: Okay,so it seems like we decided to not automatically vacuum on exit.

arhag (approved):

libraries/state_history/include/eosio/state_history/log.hpp
spoonincode: Right, we've decided not to because the size of the logs are so big that even at 100000 blocks (less than a day) you'll be waiting minutes to vacuum on consumer tier hardware.

PR #340, opened 2022-06-02
[3.2] Backporting state history plugin improvements and fixes

766C6164:
This PR is mainly targeting performance improvements with addition of threading support plus Unix socket support. A separate Issues / PR's for other backports that would be considered valuable will be created.

Backported in this PR:

- Move all network operations to another thread (8f693d3ed720056867e2bb2864ee6bde83e2570c) and all dependencies
- ported session_manager_t to make add/remove thread safe
- Unix socket support
- SHiP post send_update to main thread to avoid overwhelming the main
- Added threaded trace history/chain state history writes
- Make compatible with boost < 1.70
- Add comment why include_delta of code_object just returns false
- Pop back a delta with empty rows
- Fixed support for boolean args in unit tests (as a side bug while bringing in ut for unix socket)
- Included test for local socket connection and added support o ship_client to connect via domain socket


766C6164:
> The original PR adding ship unix socket support included adding unit test support for it. It seems like that might be something worthwhile keeping in at least one of the ship tests.

Added and it fails lol Debugging it now...

766C6164:
> > The original PR adding ship unix socket support included adding unit test support for it. It seems like that might be something worthwhile keeping in at least one of the ship tests.
> 
> Added and it fails lol Debugging it now...

Looks like ship_client needs to support unix sockets for this test to work, I'll add it

766C6164:
Proper unix_socket support implemented for ship_client, since it is closely connected to this issue, it is committed in this PR

plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -48,15 +66,23 @@ struct state_history_plugin_impl : std::enable_shared_from_this<state_history_pl
    std::optional<state_history_log> trace_log;
    std::optional<state_history_log> chain_state_log;
    bool                             trace_debug_mode = false;
-   bool                             stopping         = false;
+   std::atomic<bool>                stopping         = false;
    std::optional<scoped_connection> applied_transaction_connection;
    std::optional<scoped_connection> block_start_connection;
    std::optional<scoped_connection> accepted_block_connection;
-   string                           endpoint_address = "0.0.0.0";
+   string                           endpoint_address;
```
jgiszczak, line 73: Use a single string for the endpoint specification, either IP address or unix socket path.

plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -48,15 +66,23 @@ struct state_history_plugin_impl : std::enable_shared_from_this<state_history_pl
...
    uint16_t                         endpoint_port    = 8080;
    std::unique_ptr<tcp::acceptor>   acceptor;
```
jgiszczak, line 75: Convert `acceptor` to an `std::variant<tcp::acceptor, unixs::acceptor>` and use their common API whenever possible.

plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -100,71 +126,74 @@ struct state_history_plugin_impl : std::enable_shared_from_this<state_history_pl
...
+         ilog("incoming connection");
+         socket_stream.auto_fragment(false);
+         socket_stream.binary(true);
+         if constexpr (std::is_same_v<SocketType, tcp::socket>) {
+            socket_stream.next_layer().set_option(boost::asio::ip::tcp::no_delay(true));
+         }
+         socket_stream.next_layer().set_option(boost::asio::socket_base::send_buffer_size(1024 * 1024));
+         socket_stream.next_layer().set_option(boost::asio::socket_base::receive_buffer_size(1024 * 1024));
+        
+         socket_stream.async_accept([self = this->shared_from_this()](boost::system::error_code ec) {
```
jgiszczak, line 159: This codebase does not use explicit `this` pointers anywhere they are optional.

plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -100,71 +126,74 @@ struct state_history_plugin_impl : std::enable_shared_from_this<state_history_pl
...
-         socket_stream->binary(true);
-         socket_stream->next_layer().set_option(boost::asio::ip::tcp::no_delay(true));
-         socket_stream->next_layer().set_option(boost::asio::socket_base::send_buffer_size(1024 * 1024));
-         socket_stream->next_layer().set_option(boost::asio::socket_base::receive_buffer_size(1024 * 1024));
-         socket_stream->async_accept([self = shared_from_this()](boost::system::error_code ec) {
+      session(std::shared_ptr<state_history_plugin_impl> plugin, SocketType socket)
+          : plugin(std::move(plugin)), socket_stream(std::move(socket)) {}
+
+      void start() {
+         ilog("incoming connection");
```
jgiszczak, line 150: `fc_ilog(_log, ...` and `ilog("string")` are not interchangeable.  The explicit `_log` argument is a different file from the one used by the `ilog()` variant which accepts a string as the first argument.

plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -100,71 +126,74 @@ struct state_history_plugin_impl : std::enable_shared_from_this<state_history_pl
...
-
-      template <typename T>
-      void send(T obj) {
-         send_queue.push_back(fc::raw::pack(state_result{std::move(obj)}));
-         send();
-      }
+      
 
-      void send() {
+       void send() {
```
jgiszczak, line 186: Stray whitespace.  Unnecessary whitespace changes should be avoided, and the majority of the codebase uses three spaces per block indent.

plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -321,7 +368,7 @@ struct state_history_plugin_impl : std::enable_shared_from_this<state_history_pl
          app().post(priority::medium, [=]() {
             if (plugin->stopping)
                return;
-            if (ec)
+            if (ec)
```
jgiszczak, line 371: Stray whitespace.

plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -521,12 +637,21 @@ void state_history_plugin::plugin_initialize(const variables_map& options) {
...
-      my->endpoint_address = host;
-      my->endpoint_port    = std::stoi(port);
-      idump((ip_port)(host)(port));
+      auto ip_port = options.at("state-history-endpoint").as<string>();
+      if (ip_port.size()) {
+         auto port            = ip_port.substr(ip_port.find(':') + 1, ip_port.size());
+         auto host            = ip_port.substr(0, ip_port.find(':'));
+         my->endpoint_address = host;
+         my->endpoint_port    = std::stoi(port);
+         idump((ip_port)(host)(port));
```
jgiszczak, line 646: Remove debug dump, or convert to a proper `fc::dlog(_log, ...` call.

plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -557,18 +682,32 @@ void state_history_plugin::plugin_initialize(const variables_map& options) {
    FC_LOG_AND_RETHROW()
 } // state_history_plugin::plugin_initialize
 
-void state_history_plugin::plugin_startup() { 
-   handle_sighup(); // setup logging
```
jgiszczak, line 561: Setting up logging must not be lost.

tests/ship_client.cpp
```diff
@@ -1,120 +1,159 @@
...
+        eosio::convert(abidef, abi);
       }
 
-      std::cerr << "{\n   \"status\": \"set_abi\",\n   \"time\": " << time(NULL) << "\n},\n";
+      std::cerr << "{\n   \"status\": \"set_abi\",\n   \"time\": " << time(NULL)
+                << "\n},\n";
 
-      const eosio::abi_type& request_type = abi.abi_types.at("request");
-      const eosio::abi_type& result_type = abi.abi_types.at("result");
+      const eosio::abi_type &request_type = abi.abi_types.at("request");
```
jgiszczak, line 89: This codebase uses the convention of attaching type qualifiers to the type name, not the variable name.

jgiszczak (changes requested):

plugins/state_history_plugin/state_history_plugin.cpp
766C6164: Thank you, fixed.

plugins/state_history_plugin/state_history_plugin.cpp
766C6164: Fixed

plugins/state_history_plugin/state_history_plugin.cpp
766C6164: Good catch, done

plugins/state_history_plugin/state_history_plugin.cpp
766C6164: Thank you, done, not sure about msg in dlog if we need any

plugins/state_history_plugin/state_history_plugin.cpp
766C6164: Done, thank you, missed that salad after rebase to main

libraries/state_history/include/eosio/state_history/log.hpp
```diff
@@ -186,8 +233,12 @@ class state_history_log {
          header.magic = ship_magic(get_ship_version(header.magic), ship_feature_pruned_log);
 
       uint64_t pos = log.tellp();
+      
+      lock.unlock();
```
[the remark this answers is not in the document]
766C6164: Thank you, fixed (hopefully). Not sure why and whether lock needed to be given up prior to payload write (as it is done in develop-boxed branch)....

libraries/state_history/include/eosio/state_history/log.hpp
```diff
@@ -78,6 +82,15 @@ class state_history_log {
...
    chain::block_id_type    last_block_id;
 
+   std::thread                                                              thr;
+   std::atomic<bool>                                                        write_thread_has_exception = false;
+   std::exception_ptr                                                       eptr;
+   boost::asio::io_context                                                  ctx;
+   boost::asio::io_context::strand                                          work_strand{ctx};
+   boost::asio::executor_work_guard<boost::asio::io_context::executor_type> work_guard =
+       boost::asio::make_work_guard(ctx);
+   std::mutex                                                               mx;
```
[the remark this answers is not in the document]
766C6164: Done, thank you. I believe in our implementation get_entry was missing lock as well. Other methods called only from c-tor.

plugins/state_history_plugin/state_history_plugin.cpp
spoonincode: That sounds like a significant change to behavior: the my original implementation deliberately allowed both TCP & Unix socket to be available simultaneously.

tests/CMakeLists.txt
```diff
@@ -96,6 +96,10 @@ target_link_libraries(ship_streamer abieos Boost::program_options Boost::system
 
 add_test(NAME ship_test COMMAND tests/ship_test.py -v --num-clients 1 --num-requests 5000 --clean-run --dump-error-detail WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
 set_property(TEST ship_test PROPERTY LABELS nonparallelizable_tests)
+if(NOT WIN32)
```
spoonincode, line 99: This check can be removed: modern boost & windows supports unix sockets fine and all the original `BOOST_ASIO_HAS_LOCAL_SOCKETS` checks have already been removed

libraries/state_history/include/eosio/state_history/log.hpp
```diff
@@ -119,9 +132,37 @@ class state_history_log {
             vacuum();
          }
       }
+
+       thr = std::thread([this] {
```
spoonincode, line 136: An `eosio::chain::named_thread_pool` (even with a size of 1) may allow you to remove some of this boilerplate. Not critical, judgement up to you, just pointing it out.

spoonincode: eh the whole `write_thread_has_exception` thing may torpedo that, dunno. Still worth thinking over.

plugins/state_history_plugin/state_history_plugin.cpp
766C6164: Removing  them breaks compilation, must be there is a workaround I cant see... Thank you!

tests/ship_client.cpp
766C6164: Not sure what should be done here, we have similar code in ship_streamer as well

tests/CMakeLists.txt
766C6164: Done, thank you!

tests/ship_client.cpp
jgiszczak: `eosio::abi_type& request_type` rather than `eosio::abi_type &request_type`.  I personally am most familiar/comfortable with the latter since I started out writing C which used that convention for pointers, but this repository uses the former.

plugins/state_history_plugin/state_history_plugin.cpp
jgiszczak: Oh I didn't realize they were intended to operate simultaneously.  In that case changes made to address my comments about consolidating endpoint strings and the acceptors in a variant should be backed out.

tests/ship_client.cpp
766C6164: Ah, I see, got it! I thought something is wrong with the function used :) I like it more the C way as well. I screwed up formatting of this file, so I did clang-format it, and it does it...  Fixed.

plugins/state_history_plugin/state_history_plugin.cpp
766C6164: I knew that they can be active at the same time, but missed it during this refactoring... I adapted new pattern to both acceptors active case. Looks still cleaner to me then initial implementation with two separate methods/etc. If you think it is overkill, lmk I will roll it back :)

tests/ship_client.cpp
```diff
@@ -1,65 +1,90 @@
...
+    statehistory_port = socket_address.substr(colon + 1);
+  }
+
+  std::cerr << "[\n{\n   \"status\": \"construct\",\n   \"time\": "
+            << time(NULL) << "\n},\n";
+
+  try {
+    auto run = [&]<typename SocketStream>(SocketStream &stream) {
       {
-         boost::beast::flat_buffer abi_buffer;
```
jgiszczak, line 54: Looks like the reformatter's activities are still present in this file, both here and below.

libraries/state_history/include/eosio/state_history/log.hpp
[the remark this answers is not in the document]
spoonincode: In general, the mutexes are a smell. Proper leveraging of asio should eliminate the need for them. I'm not sure if it's worth heavy refactoring of what is intended to be backport PR though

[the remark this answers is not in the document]
766C6164: Removed  unlock

tests/ship_client.cpp
766C6164: I adjusted clang formatting options and re-formatted with 3 spaces, refs/ptrs on the left, etc. Seems to be close to what we need. Contemts of .clang-format:
`BasedOnStyle: LLVM
UseTab: Never
IndentWidth: 3
PointerAlignment: Left`

jgiszczak: Clang format seems to have separate knobs for IndentWidth in methods vs functions.  `main()` got reindented to 2 space indent for some reason.  I think something else from the LLVM style needs to be overridden.

I'm not wild about allowing clang format to touch the file because of the long swath in the middle of it that was mis-indented at 5 spaces instead of 6.  Github diff isn't good about highlighting actual code differences as opposed to mere whitespace changes.

766C6164: Re-formatted manually. I will revisit clang-format config later, out of scope of this PR. Main issue is with wrapping long lines and couple more less harmful.

jgiszczak (approved):

Issue #374, opened 2022-06-14
Investigate Nodeos "End of file" with SHiP Client Test


ClaytonCalabrese:
When using the SHiP Client test, a "End of file" error is reported in nodeos. The reason for this error is to be investigated and determined whether it is spurious, or whether it indicates an actual issue with how SHiP testing is done.

ClaytonCalabrese:
The "End of file" is the return errorcode from boost when the connection (socket) is closed. This shouldn't be interpreted as an error and so it should be suppressed in on_fail.

heifner:
Resolved by #377

PR #377, opened 2022-06-14
Suppress "End of file" error reporting in SHiP

ClaytonCalabrese:
### SHiP is incorrectly reporting an "End of file" error from async_read, which in actuality isn't an error
As such, this error is now being suppressed. It will still be logged, but not as an error. This "End of file" came through when a connection (socket) was being closed.

### DIFF
--- a/plugins/state_history_plugin/state_history_plugin.cpp
+++ b/plugins/state_history_plugin/state_history_plugin.cpp
@@ -288,7 +288,7 @@ struct state_history_plugin_impl : std::enable_shared_from_this<state_history_pl
 
       void on_fail(boost::system::error_code ec, const char* what) {
          try {
-            if (ec.message() == "End of file") {
+            if (ec == boost::asio::error::eof) {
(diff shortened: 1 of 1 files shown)


ClaytonCalabrese:
Resolves https://github.com/eosnetworkfoundation/mandel/issues/374

plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -288,7 +288,11 @@ struct state_history_plugin_impl : std::enable_shared_from_this<state_history_pl
 
       void on_fail(boost::system::error_code ec, const char* what) {
          try {
-            elog("${w}: ${m}", ("w", what)("m", ec.message()));
+            if (ec.message() == "End of file") {
```
heifner, line 291: Should be `ec == boost::asio::error::eof`

heifner (changes requested):

heifner (approved):

Issue #367, opened 2022-06-13
use sh instead of bash for cmake unittests


ClaytonCalabrese:
No need to use bash for this, sh will suffice. Makes this work on distros that don't ship with bash installed by default (Alpine), and bash's days may be numbered on macos anyways.

https://github.com/EOSIO/eos/pull/8503

Issue #356, opened 2022-06-08
Build script bash path


eosusa:
On a fresh Ubuntu 18.04 install I had to update both install_prereqs.sh and pinned_build.sh to have /bin/bash for the path instead of /usr/bin/bash as they failed to launch

matthewdarwin:
Do we need bash at all?  What about just using /bin/sh?

spoonincode:
Resolved in 3.1rc2 :+1:

PR #378, opened 2022-06-14
use sh instead of bash for cmake unittests

ClaytonCalabrese:
> No need to use bash for this, sh will suffice. Makes this work on distros that don't ship with bash installed by default (Alpine), and bash's days may be numbered on macos anyways.

### DIFF
--- a/scripts/install_deps.sh
+++ b/scripts/install_deps.sh
@@ -1,4 +1,4 @@
-#! /bin/sh
+#! /usr/bin/bash
 
 apt-get update
 apt-get update --fix-missing
--- a/scripts/install_deps.sh
+++ b/scripts/install_deps.sh
@@ -1,4 +1,4 @@
-#! /usr/bin/bash
+#! /bin/sh
 
 apt-get update
 apt-get update --fix-missing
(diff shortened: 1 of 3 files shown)


ClaytonCalabrese:
Resolves https://github.com/eosnetworkfoundation/mandel/issues/367

heifner (approved):

ClaytonCalabrese:
Resolves https://github.com/eosnetworkfoundation/mandel/issues/356

jgiszczak:
`bash` is invoked by Cluster.py to run `bios_boot.sh`.  It's used by eosio-launcher in its main.cpp for the same reason.  It's also the explicit shell used for each of:

- `tests/p2p_tests/dawn_515/test.sh`
- 'tests/db_modes_test.sh`
- 'tests/full-version-label.sh`

jgiszczak (changes requested):
`bash` is invoked by Cluster.py to run `bios_boot.sh`.  It's used by eosio-launcher in its main.cpp for the same reason.  It's also the explicit shell used for each of:

- `tests/p2p_tests/dawn_515/test.sh`
- 'tests/db_modes_test.sh`
- 'tests/full-version-label.sh`

heifner (approved):
Lets work @jgiszczak changes in a different PR.

jgiszczak (approved):

Issue #364, opened 2022-06-10
pinned builds "build" directory is not relative to the point where the build starts


matthewdarwin:
If I run `scripts/pinned_build.sh /local/lib/mandel-3.1 build 8`, I expect the build files in `build` directory relative to the current directory where I start the build.  However, the builds end up in `/local/lib/mandel-3.1/build`

linh2931:
The fix was pulled into main by https://github.com/eosnetworkfoundation/mandel/pull/494. Both release/3.1.x and main have the fix.

Issue #363, opened 2022-06-10
scripts/pinned_build.sh always prints "Currently only supporting Ubuntu based builds. Proceed at your own risk."


matthewdarwin:
scripts/pinned_build.sh always prints "Currently only supporting Ubuntu based builds. Proceed at your own risk."  Even on Ubuntu.

Code is below, but $NAME is never set before trying to check it.

```
#! /usr/bin/bash

echo "Mandel Pinned Build"

if [[ $NAME != "Ubuntu" ]]
   then
      echo "Currently only supporting Ubuntu based builds. Proceed at your own risk."
fi
```

PR #382, opened 2022-06-14
Print out warning only if platform is not Ubuntu and make build directory relative to where build starts

linh2931:
https://github.com/eosnetworkfoundation/mandel/issues/363 reports `pinned_build.sh` always prints "Currently only supporting Ubuntu based builds. Proceed at your own risk.", even on Ubuntu platform.

https://github.com/eosnetworkfoundation/mandel/issues/364 reports the build directory is made relative to the dependency directory; it should have been relative to the directory where the script starts.

This PR fixes those two issues.

### DIFF
--- a/scripts/pinned_build.sh
+++ b/scripts/pinned_build.sh
@@ -2,7 +2,7 @@
 
 echo "Mandel Pinned Build"
 
-if [[ $NAME != "Ubuntu" ]]
+if [[ $(uname -v) != *Ubuntu* ]]
    then
       echo "Currently only supporting Ubuntu based builds. Proceed at your own risk."
 fi
@@ -112,6 +112,8 @@ pushdir ${DEP_DIR}
 install_clang ${DEP_DIR}/clang-${CLANG_VER}
 install_llvm ${DEP_DIR}/llvm-${LLVM_VER}
 install_boost ${DEP_DIR}/boost_${BOOST_VER//\./_}
+
+popdir ${SCRIPT_DIR}
 
 pushdir ${MANDEL_DIR}
 


scripts/pinned_build.sh
```diff
@@ -2,7 +2,7 @@
 
 echo "Mandel Pinned Build"
 
-if [[ $NAME != "Ubuntu" ]]
+if [[ $(uname -v) != *Ubuntu* ]]
```
spoonincode, line 5: fwiw this may not yield expected behavior when running ubuntu in a container. Also, how safe is it for `uname -v` to not be in quotes here. Locally I see spaces in that
```
$ uname -v
#1 SMP PREEMPT_DYNAMIC Thu, 09 Jun 2022 16:14:10 +0000
```

matthewdarwin: I was expecting something more like:

```
#!/bin/sh

if [ -e /etc/os-release ]
then
        . /etc/os-release
fi

echo $ID
```

matthewdarwin: This will not work.  I am running a Ubuntu lxc container on a debian host.  `uname -a shows` Debian.  It is not debian (only the kernel is Debian).

linh2931: Thanks @matthewdarwin. I have changed to use os-release and provide specific reason when a build is not support.

spoonincode:
This looks good but we'll want the change in 3.1 so based on our branch strategy this change needs to be PRed to the 3.1 branch first. (sorry for the unfortunate timing!) You'll need to open a new PR with these changes based on the 3.1 release branch.

linh2931:
Do `release/3.1.x` PR https://github.com/eosnetworkfoundation/mandel/pull/487 first. And then merge that to main. Close this one.

Issue #386, opened 2022-06-15
Backport: plugin_http_api_test - fix timing issue


heifner:
https://github.com/EOSIO/eos/pull/10103

Fixes flakey test.

heifner:
Resolved by #387

PR #387, opened 2022-06-15
plugin_http_api_test - fix timing issue

heifner:
Backports: https://github.com/EOSIO/eos/pull/10103

From https://github.com/EOSIO/eos/pull/10103:
Consider the following scenario:

The overall plan is to activate a total of 17 protocol features.

The first 5 protocol features are activated and added to some block (let's say block 18).
While still on block 18, the remaining 12 protocol features transactions are executed (and they will be added to block 19, and the features will be active on block 20), and the check for waitForHeadToAdvance also gets executed while we are still on this block 18.
When the chain advances to block 19, the waitForHeadToAdvance method returns, but only 5 protocol features are really activated, leading to a state on the chain where we return from this method, but the protocol features are in realityy still not activated, leading to some failure on some further asserts that assume that this is the case after we return from preactivateAllBuiltinProtocolFeature.
Solution:

To handle this edge case, we wait 2 blocks, instead of 1, to make sure that all the transactions for protocol feature activations are executed and activated, also modify the timeout of the check for head advance to be proportional to the number of blocks we want to wait to advance.

Resoves #386

### DIFF
--- a/tests/Node.py
+++ b/tests/Node.py
@@ -1394,7 +1394,7 @@ class Node(object):
             timeout = 6 + blocksToAdvance / 2
         def isHeadAdvancing():
             return self.getHeadBlockNum() >= currentHead + blocksToAdvance
-        return Utils.waitForTruth(isHeadAdvancing, timeout)
+        return Utils.waitForBool(isHeadAdvancing, timeout)
 
     def waitForLibToAdvance(self, timeout=30):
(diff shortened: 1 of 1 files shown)


linh2931 (approved):
Backport the PR in the original form. A new issue https://github.com/eosnetworkfoundation/mandel/issues/406 was created to revisit activating the hardcoded list of protocol features.

PR #395, opened 2022-06-15
[3.2] log integrity hash on start/stop with integrity-hash-on-start & integrity-hash-on-stop options

spoonincode:
Unequivocally logging the integrity hash after loading a snapshot increases start up time from a snapshot by nearly 33%. Most users probably don't even notice the log; let's not always do that.

Log integrity hash on start up and shut down via new `integrity-hash-on-start` & `integrity-hash-on-stop` options.

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -247,6 +247,7 @@ struct controller_impl {
    named_thread_pool               thread_pool;
    platform_timer                  timer;
    deep_mind_handler*              deep_mind_logger = nullptr;
+   bool                            okay_to_print_integrity_hash_on_exit = false;
 #if defined(EOSIO_EOS_VM_RUNTIME_ENABLED) || defined(EOSIO_EOS_VM_JIT_RUNTIME_ENABLED)
    vm::wasm_allocator               wasm_alloc;
 #endif
@@ -696,6 +697,7 @@ struct controller_impl {
 
       if( conf.integrity_hash_on_start )
          ilog( "chain database started with hash: ${hash}", ("hash", calculate_integrity_hash()) );
+      okay_to_print_integrity_hash_on_stop = true;
 
       replay( check_shutdown ); // replay any irreversible and reversible blocks ahead of current head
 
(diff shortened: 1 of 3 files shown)


libraries/chain/controller.cpp
```diff
@@ -723,6 +724,9 @@ struct controller_impl {
    ~controller_impl() {
       thread_pool.stop();
       pending.reset();
+      //using the presence of a row in datebase_header index to indicate controller had a successful startup()
```
spoonincode, line 727: eh something better?

heifner: Move the chain_plugin where you know if started successfully?

matthewdarwin:
Also for 3.1?

libraries/chain/controller.cpp
spoonincode: I can't log the startup in chain_plugin because I need to to do the log before replay but after snapshot is loaded. So, it felt weird to do the startup log in controller but the shutdown log in chain_plugin.

I'm not sure doing it in chain_plugin really helps either since there are so many exceptions that can start unrolling the world at any time.

Actually I'm not really even sure this here works well enough: what if an exception is thrown half way through loading a snapshot. Then once everything unrolls the dtor is fired and if datebase_header made it then it does a snapshot with a half initialized state? That's what I want to avoid.

But I also wanted to avoid adding a `bool yep_controller_started_up`. It's kinda wild there isn't a clear indicator anywhere :thinking:

heifner: In `chain_plugin::plugin_startup` you could add a `catch(...)` after the `database_guard_exception` catch and add a `my->chain.reset();`. Then in `chain_plugin::plugin_shutdown()` you could check `my->chain` to know if it is valid, if so then log the integrity.

spoonincode: I capitulated and just added a bool :confused:

spoonincode:
> Also for 3.1?

Certainly a bit of a stretch by the typical playbook but maybe we can have that discussion since it's such a few line difference.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -373,6 +373,8 @@ void chain_plugin::set_program_options(options_description& cli, options_descrip
          ("terminate-at-block", bpo::value<uint32_t>()->default_value(0),
           "terminate after reaching this block number (if set to a non-zero number)")
          ("snapshot", bpo::value<bfs::path>(), "File to read Snapshot State from")
+         ("integrity-hash-on-start", bpo::bool_switch(), "Log the state integrity hash on startup")
+         ("integrity-hash-on-stop", bpo::bool_switch(), "Log the state integrity hash on shutdown")
```
heifner, line 377: Seems these could be config.ini options as well.

spoonincode: I was waiting to see if anyone would comment on that. My hunch was CLI only but since you mention it I'll move to config

heifner (approved):

Issue #470, opened 2022-06-20
Change to crypto_primitives host function signature and semantics.


larryk85:
For the host functions:

`alt_bn128_add`, `alt_bn128_mul`, `alt_bn128_pair`, `mod_exp`, `blake2_f`, and `k1_recover` the return values have changed.

Before 0 represented success and 1 represented failure.  Now, 0 still represents success but -1 represents failure.

This is done to allow for tri-state return values and specifically for `alt_bn128_pair`.

The new signature for `alt_bn128_pair` should be `int32_t alt_bn128_pair(const char* bytes, uint32_t length);` with the return value capturing whether it was a failure or true/false.

On failure it will return -1, 1 on false and 0 if true and successful.

In addition there are a couple of uses of `bool` in the new function signatures for `blake2_f` and `sha3`. These need to be changed to a fixed size type (preferably int32_t).

PR #471, opened 2022-06-20
Changes need to crypto-primitives host functions

larryk85:
See issue #470


libraries/chain/webassembly/crypto.cpp
```diff
@@ -108,94 +107,83 @@ namespace eosio { namespace chain { namespace webassembly {
...
+      return return_code::success;
    }
 
    int32_t interface::blake2_f( uint32_t rounds,
                                 span<const char> state,
                                 span<const char> message,
                                 span<const char> t0_offset,
                                 span<const char> t1_offset,
-                                bool final,
+                                int32_t final,
```
arhag, line 184: Should there be special logic to handle values other than 0 or 1 provided by the caller? What is the protocol claiming to do in that case?

spoonincode:
I think will need the deepmind test fixed up since changed protocol feature hash

larryk85:
Need to rebase to release/3.1.x and I am going to try to regen the dm logs again.

Issue #358, opened 2022-06-09
Backport Optimize push_transaction


ndcgundlach:
- Optimize push_transaction by using deque instead of vector to avoid vector having to reallocate and move transaction receipts and max transaction metas in block.
- std::deque used for boost < 1.71 now since configurable boost::container::deque is not available until boost 1.71. boost deque used if boost >= 1.71 available.
- Calculate action_receipt digest at action execution time to avoid copying of action_receipts and to bill for hash.
- Calculate transaction_receipt digest at transaction execution time to bill for hash.

Original PR: https://github.com/EOSIO/eos/pull/8147

PR #484, opened 2022-06-21
Backport push_transaction optimizations.

ndcgundlach:
Optimize push_transaction by using deque instead of vector to avoid vector having to reallocate and move transaction receipts and max transaction metas in block.
std::deque used for boost < 1.71 now since configurable boost::container::deque is not available until boost 1.71. boost deque used if boost >= 1.71 available.
Calculate action_receipt digest at action execution time to avoid copying of action_receipts and to bill for hash.
Calculate transaction_receipt digest at transaction execution time to bill for hash.

Resolves: https://github.com/eosnetworkfoundation/mandel/issues/358

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -121,16 +121,17 @@ struct building_block {
    std::optional<producer_authority_schedule> _new_pending_producer_schedule;
    vector<digest_type>                        _new_protocol_feature_activations;
    size_t                                     _num_new_protocol_features_that_have_activated = 0;
-   vector<transaction_metadata_ptr>           _pending_trx_metas;
-   vector<transaction_receipt>                _pending_trx_receipts;
-   vector<digest_type>                        _action_receipt_digests;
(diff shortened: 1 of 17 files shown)


heifner (approved):

libraries/chain/controller.cpp
```diff
@@ -121,16 +121,17 @@ struct building_block {
    std::optional<producer_authority_schedule> _new_pending_producer_schedule;
    vector<digest_type>                        _new_protocol_feature_activations;
    size_t                                     _num_new_protocol_features_that_have_activated = 0;
-   vector<transaction_metadata_ptr>           _pending_trx_metas;
-   vector<transaction_receipt>                _pending_trx_receipts;
-   vector<digest_type>                        _action_receipt_digests;
+   deque<transaction_metadata_ptr>           _pending_trx_metas;
```
spoonincode, line 124: seems like the whitespace could have been fixed up on all these

spoonincode (approved):

Issue #464, opened 2022-06-20
Backport: Send handshake for heartbeat/2 if no block


linh2931:
https://github.com/EOSIO/eos/pull/10627

Send handshake message if we have not seen or sent a block for heartbeat/2.
Always send handshake message even if it has not changed since it was last sent.

linh2931:
Resolved by https://github.com/eosnetworkfoundation/mandel/pull/495.

PR #495, opened 2022-06-22
Backport: Send handshake for heartbeat/2 if no block -- main

linh2931:
Backport https://github.com/EOSIO/eos/pull/10627. GitHub issue https://github.com/eosnetworkfoundation/mandel/issues/464.

From the original PR:
- Send handshake message if we have not seen or sent a block for heartbeat/2.
- Always send handshake message even if it has not changed since it was last sent.


plugins/net_plugin/net_plugin.cpp
```diff
@@ -1410,12 +1421,21 @@ namespace eosio {
...
    }
 
+   // called from connection strand
    void connection::enqueue_block( const signed_block_ptr& b, bool to_sync_queue) {
       peer_dlog( this, "enqueue block ${num}", ("num", b->block_num()) );
       verify_strand_in_this_thread( strand, __func__, __LINE__ );
 
       block_buffer_factory buff_factory;
       auto sb = buff_factory.get_send_buffer( b );
+      if( !sb ) {
```
heifner, line 1431: Remove this `if`, incomplete block is not possible for Mandel. Incomplete blocks were only possible for pruned blocks int `eosio/eos 2.1.x`

heifner (changes requested):

plugins/net_plugin/net_plugin.cpp
linh2931: Thanks. Removed the `if` statement.

heifner (approved):

Issue #426, opened 2022-06-17
fix libff in 3.1 tester cmakes


spoonincode:
it looks like there might be a problem with tester cmakes in 3.1rc1 based on main's #421

Issue #420, opened 2022-06-17
Fixes for mandel.cdt integration tests


oschwaldp-oci:
Currently mandel.cdt, built from source, can't build or link the integration tests.  Working out issues to make it possible to build again.

Parallel issue in mandel.cdt: https://github.com/eosnetworkfoundation/mandel.cdt/issues/23
Related PR: https://github.com/eosnetworkfoundation/mandel.cdt/pull/24

PR #502, opened 2022-06-22
fix libff in 3.1 tester cmakes

oschwaldp-oci:
Fix issues seen with libff when trying to use mandel with mandel.cdt.

Update CMakeLists.txt to create a parallel mandel-config.cmake for use until mandel goes through renaming.

Update path to libff in fc libraries when using build from src.

Add install directions for current mandel use with mandel.cdt (or others).

Resolves: https://github.com/eosnetworkfoundation/mandel/issues/426
Resolves: https://github.com/eosnetworkfoundation/mandel/issues/420

Supersedes: https://github.com/eosnetworkfoundation/mandel/pull/421

Relates to mandel.cdt issue/PR:
https://github.com/eosnetworkfoundation/mandel.cdt/issues/25
https://github.com/eosnetworkfoundation/mandel.cdt/pull/26

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,18 @@ We support the following CMake options:
 
 I highly recommend the ccache options. They don't speed up the first clean build, but they speed up future clean builds after the first build.
 
+### Installing for use with mandel.cdt (or others)
+
+```
+git submodule update --init --recursive
+mkdir build
+cd build
+cmake -DCMAKE_BUILD_TYPE=Release ..
+make -j $(nproc)
+make install
+cmake --install . --component dev
+```
+
 ### Running tests
 
 ```
--- a/README.md
+++ b/README.md
@@ -71,9 +71,7 @@ git submodule update --init --recursive
 mkdir build
 cd build
 cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_INSTALL_PREFIX=/usr/local ..
-make -j $(nproc)
-make install
-cmake --install . --component dev
+make -j $(nproc) dev-install
 ```
 
 ### Running tests
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ I highly recommend the ccache options. They don't speed up the first clean build
 git submodule update --init --recursive
 mkdir build
 cd build
-cmake -DCMAKE_BUILD_TYPE=Release ..
+cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_INSTALL_PREFIX=/usr/local ..
 make -j $(nproc)
 make install
 cmake --install . --component dev
--- a/README.md
+++ b/README.md
@@ -64,13 +64,17 @@ We support the following CMake options:
 
 I highly recommend the ccache options. They don't speed up the first clean build, but they speed up future clean builds after the first build.
 
-### Installing for use with mandel.cdt (or others)
+### Installing mandel dev for use with mandel.cdt (or others)
+
+The following instructions will install mandel and the development packages.  Installing is not strictly necessary. Instructions are provided, for example, in `mandel.cdt` for development and testing against a `mandel` source build directory (which may be preferable).
+
+**Warning:** It is highly recommended to select a `CMAKE_INSTALL_PREFIX` location different than the default `/usr/local`.  Here, `$HOME/mandeldev` is used as an example that should be safe for most dev systems, but each developer should determine if another location is better suited to their specific needs.
 
 ```
 git submodule update --init --recursive
 mkdir build
 cd build
-cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_INSTALL_PREFIX=/usr/local ..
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -64,6 +64,18 @@ We support the following CMake options:
...
+### Installing for use with mandel.cdt (or others)
+
+```
+git submodule update --init --recursive
+mkdir build
+cd build
+cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_INSTALL_PREFIX=/usr/local ..
+make -j $(nproc)
+make install
+cmake --install . --component dev
```
spoonincode, line 76: make -j $(nproc) dev-install will combine all 3 of these lines

README.md
```diff
@@ -64,6 +64,18 @@ We support the following CMake options:
 
 I highly recommend the ccache options. They don't speed up the first clean build, but they speed up future clean builds after the first build.
 
+### Installing for use with mandel.cdt (or others)
+
+```
+git submodule update --init --recursive
+mkdir build
+cd build
+cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_INSTALL_PREFIX=/usr/local ..
```
spoonincode, line 73: We don't want to encourage users to install dev components in to /usr/local because it causes conflicts. Will need to use something else (idk what we've been indecisive; maybe $HOME/mandeldev :shrug: ).

spoonincode:
Okay, we definitely do not want to encourage /usr/local though. It's a disaster. I'd even say there needs to be something in the README warning users to place it elsewhere since /usr/local is the default

larryk85:
> > Okay, we definitely do not want to encourage /usr/local though. It's a disaster. I'd even say there needs to be something in the README warning users to place it elsewhere since /usr/local is the default
> 
> That sounds like a good idea. Is there a specific location you'd like to use? Do you like your previous suggestion of `$HOME/mandeldev` enough to move forward with that? I can make something generic and then connect the dots between mandel and mandel.cdt README instructions.

I would recommend making this one generic so $HOME/mandeldev is fine with me.

For the most part in the future devs should have the dev package available so this instruction would be for outliers anyway.

README.md
oschwaldp-oci: Update: 2d05d35

spoonincode:
Yeah that's the thing, you don't need to install it to use it with cdt, so do we really want to encourage installing it in the first place? (With a marque entry in the README no less)

The existing mandel-contracts documentation [already is written ](https://github.com/eosnetworkfoundation/mandel-contracts#building-1)as _not_ performing an install and just using the build dir. Do we really feel an alternative approach to our own suggestion in that repo needs to be prominent and encouraged here?

oschwaldp-oci:
> Yeah that's the thing, you don't need to install it to use it with cdt, so do we really want to encourage installing it in the first place? (With a marque entry in the README no less)
> 
> The existing mandel-contracts documentation [already is written ](https://github.com/eosnetworkfoundation/mandel-contracts#building-1)as _not_ performing an install and just using the build dir. Do we really feel an alternative approach to our own suggestion in that repo needs to be prominent and encouraged here?

Fair point.  How would you prefer to proceed?  Simply delete the section on installation or is there a less marquee location to document this as a reference for someone specifically looking?

larryk85:
> > Yeah that's the thing, you don't need to install it to use it with cdt, so do we really want to encourage installing it in the first place? (With a marque entry in the README no less)
> > The existing mandel-contracts documentation [already is written ](https://github.com/eosnetworkfoundation/mandel-contracts#building-1)as _not_ performing an install and just using the build dir. Do we really feel an alternative approach to our own suggestion in that repo needs to be prominent and encouraged here?
> 
> Fair point. How would you prefer to proceed? Simply delete the section on installation or is there a less marquee location to document this as a reference for someone specifically looking?

 I would recommend just leaving the install step out of the README for now.  We really should be moving to using the dev packages for this use case anyway.  So, the documentation for mandel-contracts can stay as it is now, but move to listing the dev package and installation of that as the way to build it's contracts.

spoonincode:
Yeah my hunch is to shy away from even mentioning it in the README.

However, it is a behavior change compared to pre 3.1, so I could see something in the release notes long the lines of
> For users utilizing native contract unit testing, `make install` no longer installs the required header and libraries. It's recommended to utilize the headers and libraries directly from mandel's build directory, or to use the mandel-dev package. Installation of these headers and libraries can still be performed with a `make dev-install` if absolutely needed.

oschwaldp-oci:
> Yeah my hunch is to shy away from even mentioning it in the README.
> 
> However, it is a behavior change compared to pre 3.1, so I could see something in the release notes long the lines of
> 
> > For users utilizing native contract unit testing, `make install` no longer installs the required header and libraries. It's recommended to utilize the headers and libraries directly from mandel's build directory, or to use the mandel-dev package. Installation of these headers and libraries can still be performed with a `make dev-install` if absolutely needed.

Sounds like a plan. I will remove the install directions from the README.

spoonincode (approved):

larryk85 (approved):

PR #503, opened 2022-06-22
Merge changes needed for crypto primitives

larryk85:
Merge PR #498 into main


arhag:
@larryk85: For the future, don't create a PR to directly merge in a `release/*` branch into `main`. From my experience, there are problems that could arise with that approach. In addition, if you need to make changes to resolve conflicts specific to main, you will be forced to switch to another branch anyway.

So always create a branch off of main with some appropriate name and then merge the release branch into it. Then resolve any issues within that branch and create a PR from that branch to the `main` branch.

PR #504, opened 2022-06-23
[3.2] pick sane defaults for OPENSSL_ROOT_DIR on ARM macOS

spoonincode:
When doing a macos ARM build, use homebrew's default openssl location for default `OPENSSL_ROOT_DIR`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,7 +54,11 @@ set(ENABLE_MULTIVERSION_PROTOCOL_TEST FALSE CACHE BOOL "Enable nodeos multiversi
 
 # add defaults for openssl
 if(APPLE AND UNIX AND "${OPENSSL_ROOT_DIR}" STREQUAL "")
-   set(OPENSSL_ROOT_DIR "/usr/local/opt/openssl@3;/usr/local/opt/openssl@1.1")
+   if(CMAKE_SYSTEM_PROCESSOR STREQUAL "arm64")
+      set(OPENSSL_ROOT_DIR "/opt/homebrew/opt/openssl@3;/opt/homebrew/opt/openssl@1.1")
+   else()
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -54,7 +54,11 @@ set(ENABLE_MULTIVERSION_PROTOCOL_TEST FALSE CACHE BOOL "Enable nodeos multiversi
 
 # add defaults for openssl
 if(APPLE AND UNIX AND "${OPENSSL_ROOT_DIR}" STREQUAL "")
-   set(OPENSSL_ROOT_DIR "/usr/local/opt/openssl@3;/usr/local/opt/openssl@1.1")
+   if(CMAKE_SYSTEM_PROCESSOR STREQUAL "arm64")
```
heifner, line 57: Shouldn't this be something like `if(APPLE AND CMAKE_SYSTEM_PROCESSOR STREQUAL "arm64")`

spoonincode: It's inside of the `if(APPLE AND UNIX AND "${OPENSSL_ROOT_DIR}" STREQUAL "")` already; so it's def only going to apply to `APPLE`. Probably could condense this some by doing something like
```
if(APPLE AND UNIX AND "${OPENSSL_ROOT_DIR}" STREQUAL "" AND CMAKE_SYSTEM_PROCESSOR STREQUAL "arm64")
   #one line for arm stuff
elseif(APPLE AND UNIX AND "${OPENSSL_ROOT_DIR}" STREQUAL "")
   #one line for x86 stuff
endif()
```
But I suppose was trying to avoid the repetition of the `APPLE AND UNIX AND "${OPENSSL_ROOT_DIR}" STREQUAL ""` part.

heifner (approved):

Issue #443, opened 2022-06-20
Backport: fix db_modes_test pipeline failure


linh2931:
https://github.com/EOSIO/eos/pull/10237
https://github.com/EOSIO/eos/pull/10333

A timing issue. Those two PRs should go together.

Issue #449, opened 2022-06-20
Backport: change db_modes_test to nonparallel


linh2931:
https://github.com/EOSIO/eos/pull/10736

PR #507, opened 2022-06-23
Fix db modes test

linh2931:
Backport https://github.com/EOSIO/eos/pull/10237, https://github.com/EOSIO/eos/pull/10333, and (https://github.com/EOSIO/eos/pull/10736: reverted)
Resolve https://github.com/eosnetworkfoundation/mandel/issues/443 and https://github.com/eosnetworkfoundation/mandel/issues/449

Issues resolved:

- Fixing a timing issue where `wait` returns 127 if `nodeos` was already killed and this non-zero return code aborted the testing.
- (Making `db_modes_test` as non-parallel: reverted)


spoonincode:
db_modes_test can be parallel. afaict it was moved to nonparallel strictly because it worked better that way on the over subscribed mac VMs b1 was running. and afaik we haven't seen any problems with db_modes_test so my preference would be to keep in parallel as designed.

spoonincode:
(it's still a lousy test that should be rewritten though :slightly_smiling_face: )

spoonincode (approved):

Issue #413, opened 2022-06-17
Backport: CLEOS -- prints non-JSON formatted error message


linh2931:
https://github.com/EOSIO/eos/pull/10059 (develop) and https://github.com/EOSIO/eos/pull/10070 (2.1.x)

In 2.0.x, `cleos` always treats error messages from `nodeos` as JSON format. But sometimes `nodes` sends non-JSON formatted error messages.  This PR prevents the throw and prints original error message when JSON parsing fails.

PR #531, opened 2022-06-24
CLEOS -- Do not throw but print non-JSON formatted error message

linh2931:
Backport https://github.com/EOSIO/eos/pull/10070
Resolve https://github.com/eosnetworkfoundation/mandel/issues/413

Before the change, `cleos` would throw when an error message was not JSON formatted. After the change, `cleos` will not throw and will print out the error message.

### DIFF
--- a/programs/cleos/httpc.cpp
+++ b/programs/cleos/httpc.cpp
@@ -253,7 +253,14 @@ namespace eosio { namespace client { namespace http {
       throw;
    }
 
-   const auto response_result = fc::json::from_string(re);
+   fc::variant response_result;
+   try {
+      response_result = fc::json::from_string(re);
+   } catch(...) {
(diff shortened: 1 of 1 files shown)


spoonincode:
Should EOSIO/eos#9349 be merged before this one? That is seemingly where the
```c++
   try {
      response_result = fc::json::from_string(re);
   } catch(...) {
      // re reported below if print_response requested
   }
```
originally comes from. (admittedly it's probably not enough to cause much headache later to take this one now, but I wasn't sure how strictly in order you were trying to go)

spoonincode (approved):

linh2931:
https://github.com/eosnetworkfoundation/mandel/pull/536 replaces this PR. Close it.

PR #532, opened 2022-06-24
misc pinned build script fixes - 3.1

### LINKED ISSUE #356: Build script bash path

spoonincode:
Resolves #356 but also other issues I've encountered when using the script on all Ubuntu versions: python needs to be installed, and on Ubuntu 18 it's an old cmake so can't use `-S` argument on cmake

### DIFF
--- a/scripts/pinned_build.sh
+++ b/scripts/pinned_build.sh
@@ -31,7 +31,6 @@ JOBS=$3
 CLANG_VER=11.0.1
 BOOST_VER=1.70.0
 LLVM_VER=7.1.0
-LIBPQXX_VER=7.2.1
 ARCH=`uname -m`
 SCRIPT_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]:-$0}"; )" &> /dev/null && pwd 2> /dev/null; )";
 START_DIR="$(pwd)"
--- a/scripts/pinned_build.sh
+++ b/scripts/pinned_build.sh
@@ -130,7 +130,7 @@ pushdir ${MANDEL_DIR}
 
 # build Mandel
 echo "Building Mandel ${SCRIPT_DIR}"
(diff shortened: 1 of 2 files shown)


scripts/pinned_build.sh
```diff
@@ -130,7 +130,7 @@ pushdir ${MANDEL_DIR}
 
 # build Mandel
 echo "Building Mandel ${SCRIPT_DIR}"
-try cmake -DCMAKE_TOOLCHAIN_FILE=${SCRIPT_DIR}/pinned_toolchain.cmake -DCMAKE_INSTALL_PREFIX=/usr/local -DCMAKE_BUILD_TYPE=Release -DCMAKE_PREFIX_PATH=${LLVM_DIR}/lib/cmake -DCMAKE_PREFIX_PATH=${BOOST_DIR}/bin -S${SCRIPT_DIR}/..
+try cmake -DCMAKE_TOOLCHAIN_FILE=${SCRIPT_DIR}/pinned_toolchain.cmake -DCMAKE_INSTALL_PREFIX=/usr/local -DCMAKE_BUILD_TYPE=Release -DCMAKE_PREFIX_PATH=${LLVM_DIR}/lib/cmake -DCMAKE_PREFIX_PATH=${BOOST_DIR}/bin ${SCRIPT_DIR}/..
```
linh2931, line 133: So will we add -S back after Ubuntu 18.04 retires?

linh2931 (approved):

scripts/pinned_build.sh
spoonincode: I'm not sure there is much utility in it how we're using it; it's more interesting when running cmake outside of both the source & build directory.

PR #535, opened 2022-06-24
Add extract-blocks option to eosio-blocklog

jgiszczak:
Adds `--extract-blocks` option to the `eosio-blocklog` tool, along with an accompanying `--output-dir` option.  This option extracts the range of blocks specified by the `-f` or `--first` block option and the `-l` or `--last` block option and writes a new block log with accompanying index file to the directory specified by the `--output-dir` option.  At least one of `--first` or `--last` must be specified.

### DIFF
--- a/programs/eosio-blocklog/CMakeLists.txt
+++ b/programs/eosio-blocklog/CMakeLists.txt
@@ -1,14 +1,22 @@
 add_executable( eosio-blocklog main.cpp )
 
+find_package(progressbar 2.0 QUIET)
+
+if(progressbar_FOUND)
+  message(STATUS "Enabling progressbar in eosio-blocklog")
+endif()
+
 if( UNIX AND NOT APPLE )
   set(rt_library rt )
 endif()
 
-target_include_directories(eosio-blocklog PUBLIC ${CMAKE_CURRENT_BINARY_DIR})
+configure_file(config.hpp.in config.hpp ESCAPE_QUOTES)
+
+target_include_directories(eosio-blocklog PUBLIC ${LIBPROGRESSBAR_INCLUDE_DIRS} ${CMAKE_CURRENT_BINARY_DIR})
 
 target_link_libraries( eosio-blocklog
         PRIVATE appbase
-        PRIVATE eosio_chain fc ${CMAKE_DL_LIBS} ${PLATFORM_SPECIFIC_LIBS} )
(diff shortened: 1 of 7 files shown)


programs/eosio-blocklog/CMakeLists.txt
```diff
@@ -1,14 +1,22 @@
 add_executable( eosio-blocklog main.cpp )
 
+find_package(progressbar 2.0 QUIET)
```
spoonincode, line 3: Adding an (albeit optional) dependency for a progress bar seems.. a bit wasteful. I can't even figure out what library provides this.. it looks like _maybe_ this https://github.com/doches/progressbar but it's lacking  `progressbar_new_percent_with_format_and_tumbler` and `progressbar_update_percent`.

And doing a google or github search for `progressbar_new_percent_with_format_and_tumbler` turns up completely empty and searching for ubuntu packages with "progressbar" files in them likewise comes up empty.

Is anyone even going to have this library? Is this better suited as a submodule or copy/pasta job instead? (the one linked above adds a dependency on to ncurses, which is something we'll want to avoid as a hard requirement though)

jgiszczak: It's my version, which I haven't pushed to Github yet.  And it's just for me, because I'll be running the thing 261 times on Mainnet and wanted to see what it's doing.

jgiszczak: It's a public library, now found [here](https://github.com/jgiszczak/progressbar).  It builds cleanly and silently without it, only using it if it's present and detected by CMake.

spoonincode: We need to be judicious on the number of dependencies we pick up and this doesn't seem like a worthwhile dependency addition. It only adds trivial functionality "just for @jgiszczak" via a niche package that realistically nobody will ever have nor will be tested.  I also would like to avoid hard dependencies on ncurses so bundling the code as-is isn't an option either.

spoonincode:
You can still have a progress indicator :shrug: 

And since this is an application that isn't run as a daemon it's possible even the "\r trick" is acceptable.

jgiszczak:
Is there anything else which needs to be done?  I'm finished with this PR.

brianjohnson5972 (approved):
One minor comment, that you can fix if you agree.

Issue #517, opened 2022-06-23
Reflection Validation Script: This functionality should be brought over but use clang instead of regex.


oschwaldp-oci:
Refer to: https://github.com/eosnetworkfoundation/mandel/issues/391

heifner:
See https://github.com/eosnetworkfoundation/mandel/issues/391

Issue #516, opened 2022-06-23
Backport: adjust validate-reflection-test to work with any directory layout


oschwaldp-oci:
https://github.com/EOSIO/eos/pull/9726

heifner:
See https://github.com/eosnetworkfoundation/mandel/issues/391

Issue #515, opened 2022-06-23
Backport: Reflection validation script


oschwaldp-oci:
https://github.com/EOSIO/eos/pull/8409

heifner:
Closing, see https://github.com/eosnetworkfoundation/mandel/pull/542

Issue #391, opened 2022-06-15
Backport: Reflection validation script


linh2931:
https://github.com/EOSIO/eos/pull/8409
https://github.com/EOSIO/eos/pull/9726

This functionality should be brought over but use clang instead of regex.

oschwaldp-oci:
Going to split this into 3 issues, one each for backports and one for the additional change from clang to regex.

heifner:
I think the `FC_REFLECT` macro should be updated to validate that all members are provided to it.
Something like: `FC_REFLECT( x_type, (m_x1)(m_x2), (m_ignored_1) )` 
This can then be enforced by the compiler using `static_assert( sizeof(type) == sizeof(reflected_type) );`

heifner:
Per @larryk85:

> I would recommend that this is just a criteria for PR review (looking at the reflects) and for structs and classes that have fields and methods, have the fields in a consistent section to help with review and updating. I understand that this puts the burden onto the review process and can still have forgotten or missed fields because it is a manual process, but this will still allow for forgotten fields anyway.

PR #542, opened 2022-06-25
[3.2] Added validation to FC_REFLECT

heifner:
Added validation to `FC_REFLECT` macro so that all members have to be listed or explicitly ignored.

Requires: https://github.com/eosnetworkfoundation/mandel-fc/pull/38
Resolves: #391 
Resolves: #515
Resolves: #516 
Resolves: #517

### DIFF
--- a/libraries/chain/include/eosio/chain/account_object.hpp
+++ b/libraries/chain/include/eosio/chain/account_object.hpp
@@ -104,7 +104,7 @@ CHAINBASE_SET_INDEX_TYPE(eosio::chain::account_object, eosio::chain::account_ind
 CHAINBASE_SET_INDEX_TYPE(eosio::chain::account_metadata_object, eosio::chain::account_metadata_index)
 CHAINBASE_SET_INDEX_TYPE(eosio::chain::account_ram_correction_object, eosio::chain::account_ram_correction_index)
 
-FC_REFLECT(eosio::chain::account_object, (name)(creation_date)(abi))
+FC_REFLECT(eosio::chain::account_object, (name)(creation_date)(abi), (id))
 FC_REFLECT(eosio::chain::account_metadata_object, (name)(recv_sequence)(auth_sequence)(code_sequence)(abi_sequence)
-                                                  (code_hash)(last_code_update)(flags)(vm_type)(vm_version))
-FC_REFLECT(eosio::chain::account_ram_correction_object, (name)(ram_correction))
+                                                  (code_hash)(last_code_update)(flags)(vm_type)(vm_version), (id))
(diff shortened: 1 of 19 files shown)


spoonincode:
Is validating via `sizeof() `strict enough due to structure padding? For example,

```c++
struct one {
    uint32_t foo;
    uint32_t bar;
    uint64_t foobar;
};

struct two {
    uint32_t foo;
    uint64_t foobar;
};
```

the `sizeof()` these are equal. Would the FC_REFLECT check have caught if I left out `bar` in the reflection & ignore list?

heifner:
> the `sizeof()` these are equal. Would the FC_REFLECT check have caught if I left out `bar` in the reflection & ignore list?

No, I guess not. Any ideas for making it better? Do you think that corner case is enough to not do this?

heifner:
> No, I guess not. Any ideas for making it better? Do you think that corner case is enough to not do this?

I should be able to use `sizeof(foo) + sizeof(bar) +sizeof(foobar) == sizeof(foo) + sizeof(foobar)` instead.

heifner:
The only way I found to count struct fields or determine fields of a struct requires heavy meta-template programming that I don't think is worth pulling in for this small use case.

I tried to derive a struct from the reflected TYPE and add `__attribute__ ((packed))` but that is not portable. clang doesn't pack the derived from struct in that case. I was hoping to compare the packed sizes of the generated struct and the original struct.

I'm currently out of ideas on making this better. However, I do think it adds some level of validation that is useful.

spoonincode:
eh, personally I'm worried about the false sense of security it brings. But yes there is a different outlook that anything is better than nothing. I defer to @arhag or @larryk85

larryk85:
> eh, personally I'm worried about the false sense of security it brings. But yes there is a different outlook that anything is better than nothing. I defer to @arhag or @larryk85



> eh, personally I'm worried about the false sense of security it brings. But yes there is a different outlook that anything is better than nothing. I defer to @arhag or @larryk85

I think this also has issues with no_unique_address attributes and zero sized fields for tags. If you list all the fields it will fail because the size is greater than.

I know we currently aren't using tags in this way but it would limit us in the future.
We could try this until we need to do something else.

I would recommend that this is just a criteria for PR review (looking at the reflects) and for structs and classes that have fields and methods, have the fields in a consistent section to help with review and updating.  I understand that this puts the burden onto the review process and can still have forgotten or missed fields because it is a manual process, but this will still allow for forgotten fields anyway.

heifner:
Check via peer review sounds ok to me until we get proper C++ reflection.

Issue #425, opened 2022-06-17
Backport: Better error messaging on names with invalid characters


linh2931:
https://github.com/EOSIO/eos/pull/9186

https://github.com/EOSIO/eos/issues/8971: Invalid characters are "detected" and an error is thrown however, it is thrown by the canonical check and presents an incorrect cause of the issue for instance

linh2931:
Won't backport this, as the check is too restricted. From the PR review https://github.com/eosnetworkfoundation/mandel/pull/548 by @tbfleming "The RPC API currently accepts names using the permissive rules; this will cause it to reject them, possibly breaking existing clients. e.g. `"foo.bar"`, `"foo_bar"`, `"foo bar"`, and `"foo.bar......."` are currently accepted as synonyms for the same thing."

PR #548, opened 2022-06-27
[3.2] Better error messaging on eosio::name with invalid characters

linh2931:
Backport https://github.com/EOSIO/eos/pull/9186. Resolve https://github.com/eosnetworkfoundation/mandel/issues/425

Please see the detailed problem description in the original issue https://github.com/EOSIO/eos/issues/8971. Here is a summary.

In `name.hpp`, `string_to_uint64_t(...)` is used to convert a `std::string_view` at runtime to the proper value for an `eosio::name`. However, `string_to_uint64_t(...)` does not validate that the input characters are part of the valid character set for `eosio::name` s, instead any invalid character is turned into a 0: Invalid characters are "detected" later and an error is thrown, however, it is thrown by the canonical check and presents an incorrect cause of the issue for instance.

### DIFF
--- a/libraries/chain/include/eosio/chain/name.hpp
+++ b/libraries/chain/include/eosio/chain/name.hpp
@@ -44,8 +44,13 @@ namespace eosio::chain {
       // The for-loop encoded up to 60 high bits into uint64 'name' variable,
       // if (strlen(str) > 12) then encode str[12] into the low (remaining)
       // 4 bits of 'name'
-      if (i < str.size() && i == 12)
-         n |= char_to_symbol(str[12]) & 0x0F;
+      if (i < str.size() && i == 12) {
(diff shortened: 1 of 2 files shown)


heifner (changes requested):
@tbfleming indicated this causes an incompatible change to the RPC API. https://github.com/eosnetworkfoundation/mandel/pull/25#issuecomment-1023408227

Need to understand this difference and either not apply this PR or document the difference in the RPC API. @tbfleming can you provide more information?

tbfleming:
The RPC API currently accepts names using the permissive rules; this will cause it to reject them, possibly breaking existing clients. e.g. `"foo.bar"`, `"foo_bar"`, `"foo bar"`, and `"foo.bar......."` are currently accepted as synonyms for the same thing.

PR #569, opened 2022-06-30
Update and bring to current state mandel/docs 

766C6164:
Initial revision, still work in progress


stephenpdeos (changes requested):
As discussed, we should for now revert Mandel rename from EOSIO. We will update this nomenclature once we've announced our official new name vs the stop-gap name of the repo it is in.

766C6164:
> As discussed, we should for now revert Mandel rename from EOSIO. We will update this nomenclature once we've announced our official new name vs the stop-gap name of the repo it is in.

Un-renamed these

stephenpdeos (approved):
Approved.

PR #583, opened 2022-06-30
[3.2] Use NET bill in transaction receipt during light validation mode

### LINKED ISSUE #559: Backport: Use NET bill in transaction receipt during light validation mode
> Since the NET bill is objective, a node in full validation mode is able to calculate what the NET bill for a transaction should be. Even with this change, a node in full validation mode will always compute the NET bill rather than blindly trusting the NET bill provided by the producer in the transaction receipt and it will reject the block if there is a mismatch.

However, with the changes in this PR, a node operating in a mode where the normal transaction checks are skipped (light validation mode or applying a block by a trusted producer) will now simply use the NET bill provided by the producer in the transaction receipt rather than computing the amount itself. This sets the stage for later changes that allow a light validation node to apply a block that may have had some context-free data within the block removed, which would make it impossible to calculate the correct NET bill.

This PR also adds a new unit test, resource_limits_test/light_net_validation, to test the new behavior.

Backports: https://github.com/EOSIO/eos/pull/8856

ClaytonCalabrese:
> Since the NET bill is objective, a node in full validation mode is able to calculate what the NET bill for a transaction should be. Even with this change, a node in full validation mode will always compute the NET bill rather than blindly trusting the NET bill provided by the producer in the transaction receipt and it will reject the block if there is a mismatch.
> 
> However, with the changes in this PR, a node operating in a mode where the normal transaction checks are skipped (light validation mode or applying a block by a trusted producer) will now simply use the NET bill provided by the producer in the transaction receipt rather than computing the amount itself. This sets the stage for later changes that allow a light validation node to apply a block that may have had some context-free data within the block removed, which would make it impossible to calculate the correct NET bill.
> 
> This PR also adds a new unit test, resource_limits_test/light_net_validation, to test the new behavior.

Resolves #559

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -1498,7 +1498,7 @@ struct controller_impl {
 
          try {
             if( trx->implicit ) {
-               EOS_ASSERT( !explicit_net_usage_words.valid(), transaction_exception, "NET usage cannot be explicitly set for implicit transactions" );
+               EOS_ASSERT( !explicit_net_usage_words, transaction_exception, "NET usage cannot be explicitly set for implicit transactions" );
                trx_context.init_for_implicit_trx();
                trx_context.enforce_whiteblacklist = false;
             } else {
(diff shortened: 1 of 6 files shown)


heifner (changes requested):
I don't think we want this @arhag

heifner:
This was part of prunable CFD which is not needed in Mandel.

Issue #590, opened 2022-07-02
Fix race condition in trace_api_plugin shutdown


heifner:
Failure https://github.com/eosnetworkfoundation/mandel/runs/7143297502?check_suite_focus=true points to an issue in shutdown of the `trace_api_plugin`.

`slice_directory::stop_maintenance_thread()` sets atomic bool `_maintenance_shutdown = true` and calls `notify_one()` on the `_maintenance_condition` condition variable. This is a race condition because it doesn't acquire the condition variable mutex setting up the possibility that `_maintenance_shutdown ` can be set to `true` and `notify_one` called after the while check in `slice_directory::start_maintenance_thread`  but before the `wait()` causing it to wait forever. This then blocks the `slice_directory::stop_maintenance_thread()` call of `join()` on the main thread, blocking the shutdown of all other plugins.

```
debug 2022-07-01T03:23:35.304 nodeos    net_plugin.cpp:2705           update_chain_info    ] updating chain info lib 90, head 138, fork 138
info  2022-07-01T03:23:35.304 nodeos    resource_monitor_plugi:122    plugin_shutdown      ] shutdown...
debug 2022-07-01T03:23:35.304 net-0     net_plugin.cpp:3195           operator()           ] accepted signed_block : #138 23fb4fff5f2c0077...
debug 2022-07-01T03:23:35.304 net-1     net_plugin.cpp:2152           operator()           ] ["localhost:9876 - 465c8f8" - 3 127.0.0.1:9876] bcast block 138
debug 2022-07-01T03:23:35.304 net-1     net_plugin.cpp:2173           recv_block           ] ["localhost:9876 - 465c8f8" - 3 127.0.0.1:9876] canceling wait
debug 2022-07-01T03:23:35.304 net-1     net_plugin.cpp:1993           sync_recv_block      ] ["localhost:9876 - 465c8f8" - 3 127.0.0.1:9876] got block 138
info  2022-07-01T03:23:35.304 net-1     net_plugin.cpp:1016           _close               ] ["localhost:9876 - 465c8f8" - 3 127.0.0.1:9876] closing
info  2022-07-01T03:23:35.304 net-1     net_plugin.cpp:1227           operator()           ] ["localhost:9876 - 465c8f8" - 3 127.0.0.1:9876] async write socket closed before callback
[...]

PR #592, opened 2022-07-02
[3.1] Fix race condition on trace_api_plugin shutdown

heifner:
Failure https://github.com/eosnetworkfoundation/mandel/runs/7143297502?check_suite_focus=true pointed to an issue in shutdown of the `trace_api_plugin`.

`slice_directory::stop_maintenance_thread()` sets atomic bool `_maintenance_shutdown = true` and calls `notify_one()` on the `_maintenance_condition` condition variable. This is a race condition because it doesn't acquire the condition variable mutex setting up the possibility that `_maintenance_shutdown ` can be set to `true` and `notify_one` called after the while check in `slice_directory::start_maintenance_thread`  but before the `wait()` causing it to wait forever. This then blocks the `slice_directory::stop_maintenance_thread()` call of `join()` on the main thread, blocking the shutdown of all other plugins.

Changed the logic to correctly use a mutex and condition variable around `_best_known_lib` & `_maintenance_shutdown`. Also made these two variables non-atomic since they are now only accessed via the mutex. Also release the mutex before running the `run_maintenance_tasks()` which appeared to be the original intention with the use of the local variables.

Accidentally targeted `main` on #591 so this is a cherry-pick back to `3.1`.

### DIFF
--- a/plugins/trace_api_plugin/include/eosio/trace_api/store_provider.hpp
+++ b/plugins/trace_api_plugin/include/eosio/trace_api/store_provider.hpp
@@ -251,11 +251,11 @@ namespace eosio::trace_api {
       std::optional<uint32_t> _last_compressed_slice;
       const size_t _compression_seek_point_stride;
 
-      std::atomic<uint32_t> _best_known_lib{0};
       std::mutex _maintenance_mtx;
       std::condition_variable _maintenance_condition;
       std::thread _maintenance_thread;
-      std::atomic_bool _maintenance_shutdown{false};
+      bool _maintenance_shutdown{false};
+      uint32_t _best_known_lib{0};
    };
 
    /**
--- a/plugins/trace_api_plugin/store_provider.cpp
+++ b/plugins/trace_api_plugin/store_provider.cpp
@@ -289,7 +289,10 @@ namespace eosio::trace_api {
    }
 
    void slice_directory::set_lib(uint32_t lib) {
-      _best_known_lib = lib;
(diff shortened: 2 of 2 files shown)


linh2931:
Sorry I did not notice the wrong branch on #591 either. But the line under the title of this PR still says [heifner](https://github.com/heifner) wants to merge 1 commit into [main](https://github.com/eosnetworkfoundation/mandel) from [GH-590-shutdown-race-3-1](https://github.com/eosnetworkfoundation/mandel/tree/GH-590-shutdown-race-3-1), not into `3.1`.

heifner:
> Sorry I did not notice the wrong branch on #591 either. But the line under the title of this PR still says [heifner](https://github.com/heifner) wants to merge 1 commit into [main](https://github.com/eosnetworkfoundation/mandel) from [GH-590-shutdown-race-3-1](https://github.com/eosnetworkfoundation/mandel/tree/GH-590-shutdown-race-3-1), not into `3.1`.

Can't believe I did it again. Thanks.

spoonincode (approved):

heifner:
> I think may still want to make a merge PR for this?

I accidentally did that first: https://github.com/eosnetworkfoundation/mandel/pull/591

heifner:
> Right, if you don't do a (noop) merge won't it leave a "gap" where we have a PR that wasn't merged?

Created https://github.com/eosnetworkfoundation/mandel/pull/604

Issue #384, opened 2022-06-14
Revisit logging approach in deep_mind_tests


arhag:
Given changes in https://github.com/eosnetworkfoundation/mandel/pull/383, see if there is any need for the complicated merge and comparison approach for the deep_mind logs or if it can be greatly simplified.

PR #593, opened 2022-07-03
[3.2] simplify deep_mine_tests by replacing complicated merge and match with  straight-forward file copying and comparison

linh2931:
Resolve https://github.com/eosnetworkfoundation/mandel/issues/384

https://github.com/eosnetworkfoundation/mandel/pull/383 fixed non-determinism issues with deep-mind logger by zeroing out the elapsed fields within a copy of the transaction trace before writing it out to the logs in tests. I have confirmed that fix produced deterministic logs. The complicated merge and match code in deep_mine_tests are replaced with simple file copying and comparison.


unittests/deep_mind_tests.cpp
```diff
@@ -53,363 +50,44 @@ struct deep_mind_tester : deep_mind_log_fixture, validating_tester
...
+         BOOST_TEST(false, "Mismatch at line " << i << "\n+ " << line1 << "\n- " << line2);
          return;
       }
       ++i;
    }
-   if(std::getline(file, line))
+   if(std::getline(file1, line1))
    {
-      BOOST_TEST(false, "Expected end of file at line " << i);
+      BOOST_TEST(false, "Expected end of file1 at line1 " << i);
```
swatanabe, line 86: this is a string, not a variable name.

unittests/deep_mind_tests.cpp
```diff
@@ -53,363 +50,44 @@ struct deep_mind_tester : deep_mind_log_fixture, validating_tester
...
-            for(std::ptrdiff_t j = 0; j < line_end - line_iter; ++j)
-            {
-               tab[j].push_back(add_point(j, i));
-            }
-         }
-         next(line_end - line_iter, pattern_end - pattern_iter);
-      };
-      editdiff_loop();
+   while(std::getline(input_file, s)) {
+      output_file << s << "\n";
```
swatanabe, line 60: Can use `output_file << input_file.rdbuf()` or just `fc::copy` and get rid of the streams entirely.

linh2931: Thanks for the suggestion. It further simplifies the code. I changed to use `fc:rename` as `fc:copy` does not allow destination to exist before copying.

arhag (approved):

PR #595, opened 2022-07-05
[3.2] Backport separate logging for state history plugin

ndcgundlach:
add separate logging for state history plugin


heifner (changes requested):
Should add this logger to the example `logging.json` in `programs/nodeos`.
Also update the PR description with information about how to enable this logging in SHiP.

heifner (approved):

PR #596, opened 2022-07-05
[3.1] various README.md improvements w.r.t. build & branches

### LINKED ISSUE #512: Provide documentation on how to get mandel 3.1 compiled on Ubutntu 18.04
Based on the discussion in the Jungle telegram chat, there is still interest in having mandel 3.1 running on Ubuntu 18.04 and self-compiling.

Provide some basic instructions for getting this to work in the README.md.

The instructions may be optionally prefixed with note that upgrading to 20.04 is better idea than actually trying to compile on 18.04.

spoonincode:
Various improvements and/or opinionated changes to the README.md
* Document how to do a manual build on Ubuntu 18.04 (resolves #512)
* Warn about high `-j` values (resolves #512)
* Make clear that `main` is for development not production and send users over to the releases page to get latest release & tag info; remove the release 3.0 & release 3.1 mentions as that list will just grow forever over time.
* Encourage users to grab the binaries from the release page instead of building.
* Don't use pessimistic language around distro support; be more positive and mention other distros or platforms we're treating "best-effort" even though we don't support them.
* Remove the `full-version-label-test|release-build-test|print-build-info-test` test exceptions. Day-to-day devs might not want those but when doing a build straight from the repo those tests should pass so let's just simplify this and suggest users to run all them.
* Eliminate any build options & instructions other then the bare minimum. Yes building with Ninja is nice. Yes `-DDISABLE_WASM_SPEC_TESTS=yes` speeds up build. Yes `ccache` can be a speed improvement for users who can tolerate it. But all these additions distract from the otherwise simple build instructions that are here _to get users bootstrapped easily and simply_. Perhaps these could be folded in to an "advanced build options" section or documentation or something.
* Warn users that the non-parallel tests can do bad things like SIGKILL your nodeos.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,24 +2,41 @@
 
 Home of the official [EOS Network Foundation](https://eosnetwork.com/) blockchain node software.
 
-## Repo organization
+## Repo Organization
 
-| branch                | description |
-| ------                | ----------- |
-| `main`                | Development for future releases |
-| `release/3.0.x`       | 3.0.x-* series of pre-releases before 3.1.x |
-| `release/3.1.x`       | 3.1.x* series of 3.1.0 releases |
+`main` branch is the development branch: do not use this for production. Refer to the [release page](https://github.com/eosnetworkfoundation/mandel/releases) for current information on releases, pre-releases, and obsolete releases as well as the corresponding tags for those releases.
 
-## Supported Operating Systems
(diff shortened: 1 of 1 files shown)


stephenpdeos (approved):
Approved

PR #603, opened 2022-07-05
[3.2] Update jiraIssueCreator.yml

oschwaldp-oci:
For now, remove github issue body as it will often break json format without some additional work.  For now just include the github issue link.


spoonincode (dismissed):
one sec

heifner (approved):

PR #620, opened 2022-07-07
[3.2] Backport: [statehist] Ship logs reduction / optimization

766C6164:
Had to introduce and then reduce logs here :)


plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -20,6 +20,12 @@ namespace ws = boost::beast::websocket;
 
 extern const char* const state_history_plugin_abi;
 
+// during syncing if block is older than this, reduce logging 
+const int64_t   log_reduction_minutes = 5;
```
heifner, line 24: I actually think this makes it less readable. If you are going to have a constant then make it a `fc::microseconds`.
Same goes with `log_reduction_one_per`, I would prefer just using `1000` in the code.

plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -234,6 +240,28 @@ struct state_history_plugin_impl : std::enable_shared_from_this<state_history_pl
...
+               return chain.fetch_block_by_number(block_num);
+            } catch (...) {
+               return {};
+            }
+         };
+         auto block = get_blk();
+
+         bool fresh_block = block && fc::time_point::now() - block->timestamp < fc::minutes(log_reduction_minutes);
+         if( fresh_block || (result.this_block && result.this_block->block_num % log_reduction_one_per == 0) ) {
+            ilog("pushing result "
```
heifner, line 258: If https://github.com/eosnetworkfoundation/mandel/pull/595 is merged before this, then change this to `fc_ilog`.

heifner (changes requested):

plugins/state_history_plugin/state_history_plugin.cpp
766C6164: Agreed, doubt somebody ever will be changing these. Done.

plugins/state_history_plugin/state_history_plugin.cpp
766C6164: Merged and changed

heifner (approved):

Issue #467, opened 2022-06-20
Backport: fix bug in trace_api_plugin no params returned


linh2931:
https://github.com/EOSIO/eos/pull/10449

In trace API plugin, if RPC ABIS are specified with --trace-rpc-abi, the 'params' field of the returned transaction JSON did not exist. This was caused by an exception being thrown in chain::abi_serializer, when the return_value field did not exist, it would thrown an 'Unpack exception' even though action params were serialized correctly.

Backport will also bring in tests/trace_plugin_test.py.

766C6164:
We do NOT want this. It hides abi yield errors.

PR #621, opened 2022-07-07
Backport: fix bug in trace_api_plugin no params returned #467

vladtr:
This is may not be needed, it wrapping of serializer's binary_to_variant call in try/catch as what backport suggests. Current implementation already checks for empty return_value. But can serializer throw for another reason then empty string?


heifner (changes requested):
We do NOT want this. It hides abi yield errors.

Issue #623, opened 2022-07-08
Block producer hits coma state and does not recover 


heifner:
EOSUSA testing Mandel 3.1.0-rc2 on Jungle got block producer in coma state which it would not recover from without a restart.

![image](https://user-images.githubusercontent.com/3933649/177895611-66e3268f-6a80-4bcd-9ac5-81e3d5e19414.png)

> sigh... and a simple restart of the nodeos instance and it immediately started syncing back with clions (which it had been rejecting for about the past 30 mins...)....   soooooo, even without trying hard, i think its safe to say theres still some problems withe p2p

> yeah, so i have 2 bp nodes, one peered directly to the other (BP01 & BP02).  BP01 is active signer and went into a coma about an hour ago and stayed unlinkable.  restarting it immediately allowed it to accept the real chain from clions p2p.  this other BP02 server has only a single peer, the BP01 server, but has remained in a coma not syncing blocks or rejecting them... just knowing it doesnt have the keys for signing during its scheduled rotation.  so im assuming when i restart BP02, itll begin syncing the chain from BP01 again (because BP01 is working properly again and signing/syncing), but it shows that the p2p process still isnt good at reconnecting to a peer, and now might even be worse (rarely would it ever go longer than 15mins before reconnecting, altho this is ubuntu 20 vs all others being 18)

> EOSUSA Michael | eosusa.io, [7/7/2022 6:41 AM]
![image](https://user-images.githubusercontent.com/3933649/177895813-2a122935-9f96-42a9-9842-4d7cde08af87.png)

> yep... immediately after restart, that BP02 server started catching up to the chain from BP01 like it should have 30+ mins ago... so nah, dont think the comas are fixed sadly... 😭😭😭🤬

> EOSUSA Michael | eosusa.io, [7/7/2022 6:46 AM]
And as far as i can tell, jungle was the only chain here impacted.  Normally a random coma like this is seen on pther chains too but wasnt (altho i only have clions as peer so that might have triggered it)

heifner:
```
 [xxx] Sending handshake generation 5, lib 23085437, head 23085771, id e1782c10eba9f5a6
 [xxx] Catching up with chain, our last req is 0, theirs is 23086311
 [xxx] requesting range 23085772 to 23086271
```

[...]

PR #627, opened 2022-07-09
[3.1] Fix for block producer stuck in coma state

heifner:
Backports: https://github.com/eosnetworkfoundation/mandel/pull/495 Send handshake for heartbeat/2 if no block (first commit)
* Send handshake message if we have not seen or sent a block for heartbeat/2.
* Always send handshake message even if it has not changed since it was last sent.

Additional changes in second commit:
* Main change is to always start syncing from LIB instead of HEAD.
  * See https://github.com/eosnetworkfoundation/mandel/issues/623#issuecomment-1179463090 
* Change default keepalive interval to 10s. Previous default of 32s seems rather excessive given that block time is 500ms.
* Honor keepalive interval regardless of protocol version since there is nothing that requires a particular protocol version.

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -225,7 +225,7 @@ namespace eosio {
       boost::asio::steady_timer::duration   connector_period{0};
       boost::asio::steady_timer::duration   txn_exp_period{0};
       boost::asio::steady_timer::duration   resp_expected_period{0};
-      std::chrono::milliseconds             keepalive_interval{std::chrono::milliseconds{32 * 1000}};
+      std::chrono::milliseconds             keepalive_interval{std::chrono::milliseconds{10 * 1000}};
       std::chrono::milliseconds             heartbeat_timeout{keepalive_interval * 2};
 
       int                                   max_cleanup_time_ms = 0;
@@ -392,7 +392,7 @@ namespace eosio {
    constexpr auto     def_txn_expire_wait = std::chrono::seconds(3);
    constexpr auto     def_resp_expected_wait = std::chrono::seconds(5);
    constexpr auto     def_sync_fetch_span = 100;
-   constexpr auto     def_keepalive_interval = 32000;
+   constexpr auto     def_keepalive_interval = 10000;
 
    constexpr auto     message_header_size = sizeof(uint32_t);
    constexpr uint32_t signed_block_which       = fc::get_index<net_message, signed_block>();       // see protocol net_message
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -1150,7 +1150,7 @@ namespace eosio {
 
    // called from connection strand
    void connection::check_heartbeat( tstamp current_time ) {
-      if( protocol_version >= heartbeat_interval && latest_msg_time > 0 ) {
+      if( latest_msg_time > 0 ) {
```
linh2931, line 1153: Not related to this PR. The version names of net_plugin should have a prefix of protocol or version. When I first read `protocol_version >= heartbeat_interval`, I thought `protocol_version` and `heartbeat_interval` were different units, how could they compare? Then found out `heartbeat_interval` talked about a version, not an interval value.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1148,21 +1149,31 @@ namespace eosio {
...
             no_retry = benign_other;
             if( !peer_address().empty() ) {
-               peer_wlog(this, "heartbeat timed out for peer address");
-               close(true);  // reconnect
+               fc_wlog(logger, "heartbeat timed out for peer address ${adr}", ("adr", peer_address()));
+               close(true);
             } else {
-               peer_wlog( this, "heartbeat timed out" );
+               {
+                  std::lock_guard<std::mutex> g_conn( conn_mtx );
```
linh2931, line 1161: Just wondering why mutex is required here.

linh2931 (approved):

plugins/net_plugin/net_plugin.cpp
heifner: `last_handshake_recv` is accessed across threads. Should probably log outside the mutex.

heifner: Actually this is in a connection strand call, so it should use peer_wlog.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -225,7 +225,7 @@ namespace eosio {
       boost::asio::steady_timer::duration   connector_period{0};
       boost::asio::steady_timer::duration   txn_exp_period{0};
       boost::asio::steady_timer::duration   resp_expected_period{0};
-      std::chrono::milliseconds             keepalive_interval{std::chrono::milliseconds{32 * 1000}};
+      std::chrono::milliseconds             keepalive_interval{std::chrono::milliseconds{10 * 1000}};
```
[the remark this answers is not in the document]
heifner: Done.

plugins/net_plugin/net_plugin.cpp
heifner: Some already had `proto_` so added that to all of them.

linh2931 (approved):

PR #638, opened 2022-07-10
be explict about softfloat's float128_t usage instead of boost::float128_t; fixes boost 1.80 build

spoonincode:
`abi_serializer.cpp` does a
```c++
using namespace boost;
```

Apparently something internal to boost has been reorganized and when building with boost 1.80 beta it appears this `pack_unpack<float128_t>()` is getting confused with `boost::float128_t`  (gets a `template argument deduction/substitution failed` error).

Be explicit here that we want softfloat's `float128_t`.

Sending this in to 3.1 since confusion around this type seems like a Bad Thing. I wonder if we should blanket replace all `float128_t` with `::float128_t` elsewhere though?

### DIFF
--- a/libraries/chain/abi_serializer.cpp
+++ b/libraries/chain/abi_serializer.cpp
@@ -106,7 +106,7 @@ namespace eosio { namespace chain {
       // TODO: Add proper support for floating point types. For now this is good enough.
       built_in_types.emplace("float32",                   pack_unpack<float>());
       built_in_types.emplace("float64",                   pack_unpack<double>());
-      built_in_types.emplace("float128",                  pack_unpack<float128_t>());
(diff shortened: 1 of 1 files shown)


heifner:
> I wonder if we should blanket replace all float128_t with ::float128_t elsewhere though?

I would vote, yes. If/when `float128_t` added to C++ it will likely be in namespace `std`. https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p1467r5.html . Seems like we would be better served to be explicit. Maybe even better to declare our own `using soft_float128_t = ::float128_t`

spoonincode:
hm yeah let me think that over some

spoonincode:
replaced by #702

Issue #92, opened 2022-04-09
Enhancement: allow nodeos to run in "no blocks.log" mode


matthewdarwin:
Add a new configuration option to allow nodeos to not generate any blocks.log file on disk.  Good for block producer nodes.

linh2931:
Hi @matthewdarwin, does option name `not-generate-blocks-log` sounds good? When this option is configured, I will still allow to start from a blocks.log, just not write anything into the file.

Can you also create a separate issue for your additional comment (reversible blocks) for better tracking and easier testing?

matthewdarwin:
Seems messy to have a blocks.log that is not kept up-to-date.

linh2931:
I am not aware of reversible blocks issue. Do you have an issue number?

heifner:
Yes, #180 is in `3.1.0-rc1`

linh2931:
> Seems messy to have a blocks.log that is not kept up-to-date.

OK. When `not-generate-blocks-log` is configured, `blocks.log` file will not opened and replay from it will not allowed.

spoonincode:
Another possibility for enabling "no blocks.log mode" would be via `block-log-retain-blocks = 0` (currently the setting is required to be 1 or more)

Of course, as is, that option's behavior is to prune immediately which is at odds of
> When this option is configured, I will still allow to start from a blocks.log, just not write anything into the file.

There is ongoing discussion that the `block-log-retain-blocks` option should not prune until replay is complete; maybe that aligned better with the above quote that would allow starting  with a blocks.log still (and then remove the blocks.log once replay is complete?)

linh2931:
I think `not-generate-blocks-log` conveys the intention of the nodeos operator better than `block-log-retain-blocks` being 0. It also avoids the call to block_log's `append`, which saves it the trouble from deciding not to write to the blocks.log if `block-log-retain-blocks` is 0.

spoonincode:
I'm not suggesting leaning on the existing pruning code to support no blocks.log mode; as you mention the logic should be different (i.e. no file, no `append()` calls, etc). I'm suggesting to at least consider overloading the option instead of adding a new option. After all, someone who wants no block log wants to _retain zero blocks_.

linh2931:
Thanks Matt for clarification. It is clear to me now. I agree to use `block-log-retain-blocks = 0`, so I don't add one more new option to the code and the users.

matthewdarwin:
[...]

PR #643, opened 2022-07-11
[3.2] Enhancement: Add capability of not genereating block log by setting option block-log-retain-blocks to 0

linh2931:
Resolve https://github.com/eosnetworkfoundation/mandel/issues/92.

`block-log-retain-blocks` option is modified to allow being set to 0, which tells `nodeos` not to generate blocks.log.

Now `block-log-retain-blocks` description looks like
```
  --block-log-retain-blocks arg         If set to greater than 0, periodically
                                        prune the block log to store only
                                        configured number of most recent
                                        blocks.
                                        If set to 0, no blocks are be written
                                        to the block log; block log file is
                                        removed after startup.
```

New unit tests and basic integration tests are added for this enhancement.


libraries/chain/block_log.cpp
```diff
@@ -628,6 +661,12 @@ namespace eosio { namespace chain {
    signed_block_ptr block_log::read_block_by_num(uint32_t block_num)const {
       try {
          signed_block_ptr b;
+
+         if (my->not_generate_block_log) {
+            // No blocks exist. Avoid cascading failures if going further.
+            return b;
+         }
```
heifner, line 668: There are a number of public methods of `block_log` that should also check this flag. Including: read_block, read_block_header, read_block_by_num, read_block_id_by_num, read_block_by_id, get_block_pos, read_head, flush, etc. These should all be updated to do the correct thing when in `not_generate_block_log` mode.

heifner: Ideally we should  have unittests around this new functionality so we can show all these work in the new mode.

libraries/chain/controller.cpp
```diff
@@ -570,6 +570,9 @@ struct controller_impl {
          shutdown();
       }
 
+      if (conf.prune_config && conf.prune_config->prune_blocks == 0) {
+         blog.remove();
+      }
```
heifner, line 575: Maybe it is fine to blow away the blocklog in these cases, but I wonder if we should require something like --delete-all to be passed for that to happen automatically. I would hate to accidently delete my 3TB blocklog. The config description does indicate it deletes the blocklog when done, so I guess this is ok.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -868,7 +868,17 @@ void chain_plugin::plugin_initialize(const variables_map& options) {
       if(options.count( "block-log-retain-blocks" )) {
          my->chain_config->prune_config.emplace();
          my->chain_config->prune_config->prune_blocks = options.at( "block-log-retain-blocks" ).as<uint32_t>();
-         EOS_ASSERT(my->chain_config->prune_config->prune_blocks, plugin_config_exception, "block-log-retain-blocks cannot be 0");
+
+         if ( my->chain_config->prune_config->prune_blocks == 0 ) {
+            // clear out empty blocks.log. otherwise block_log::extract_genesis_state
+            // will return version 0 which asserts.
```
heifner, line 874: How do you get into a state with a blocks.log of 0 size?

heifner (changes requested):

tests/block_log_retain_blocks_test.py
```diff
@@ -0,0 +1,101 @@
...
+
+def is_in_file(string, filename):
+   with open(filename) as myfile:
+      if string in myfile.read():
+         return True
+      else:
+         return False
+
+def start_and_stop_nodeos(retain_blocks=0, produced_block=10):
+   cmd=f'programs/nodeos/nodeos -e -p eosio --plugin eosio::producer_api_plugin --plugin eosio::producer_plugin --plugin eosio::chain_api_plugin --plugin eosio::chain_plugin --data-dir {data_dir} --block-log-retain-blocks {retain_blocks}'
```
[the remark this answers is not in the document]
spoonincode: Also consider disabling p2p & http endpoints so it can be a parallelizable test. Can Node/Cluster do that for us @heifner ?

libraries/chain/block_log.cpp
```diff
@@ -49,14 +49,22 @@ namespace eosio { namespace chain {
...
-                  EOS_ASSERT(__builtin_popcount(prune_config->prune_threshold) == 1, block_log_exception, "block log prune threshold must be power of 2");
-                  //switch this over to the mask that will be used
-                  prune_config->prune_threshold = ~(prune_config->prune_threshold-1);
+                  if (prune_config->prune_blocks == 0 ) {
+                     // not to generate blocks.log
+                     // disable prune log handling by resetting prune_config
+                     prune_config.reset();
+                     not_generate_block_log = true;
+                  } else {
+                     EOS_ASSERT(prune_config->prune_blocks, block_log_exception, "block log prune configuration requires at least one block");
```
spoonincode, line 63: this assert can be removed now

libraries/chain/controller.cpp
spoonincode: Doesn't seem any more dangerous than accidentally doing an existing passively allowed `block-log-retain-blocks = 1`. Not sure where the threshold should be for a `--i-really-mean-it` like option

heifner: The description does indicate it deletes the blocklog, so that is good enough for me at least.

tests/block_log_retain_blocks_test.py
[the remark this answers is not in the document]
heifner: You can start a cluster of 1 with `extraNodeosArgs` set with the `--p2p-listen-endpoint`. See almost any of the python tests.
One of the items I hope we can start on soon is making all the python integration tests able to run in parallel. By using Cluster.py when that work is done this test would pick put that ability along with it.

libraries/chain/block_log.cpp
```diff
@@ -398,6 +415,11 @@ namespace eosio { namespace chain {
       FC_LOG_AND_RETHROW()
    }
 
+   void detail::block_log_impl::update_head(const signed_block_ptr& b) {
+      head = b;
+      head_id = b->calculate_id();
+   }
```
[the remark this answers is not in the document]
linh2931: Thanks. I will use common function `update_head` in `append` and `open`.

libraries/chain/block_log.cpp
linh2931: Thanks. I missed that assert.

plugins/chain_plugin/chain_plugin.cpp
linh2931: I intentionally tested with blocks.log of 0 size. Probably it is not a valid test. Will revert this code.

libraries/chain/block_log.cpp
```diff
@@ -631,6 +681,13 @@ namespace eosio { namespace chain {
    signed_block_ptr block_log::read_block_by_num(uint32_t block_num)const {
       try {
          signed_block_ptr b;
+
+         if (my->not_generate_block_log) {
+            // No blocks exist. Avoid cascading failures if going further.
+            ilog("Not possible to read_block_by_num in no blocks.log mode (block-log-retain-blocks=0)");
```
heifner, line 687: Seems rather chatty. Consider asserting or just returning null with no log.

libraries/chain/block_log.cpp
```diff
@@ -643,6 +700,10 @@ namespace eosio { namespace chain {
 
    block_id_type block_log::read_block_id_by_num(uint32_t block_num)const {
       try {
+         if (my->not_generate_block_log) {
+            ilog("Not possible to read_block_id_by_num in no blocks.log mode (block-log-retain-blocks=0)");
```
heifner, line 704: All these "error" ilogs seem rather chatty. Consider just return nulls or asserting if they should never be called in no block log mode.

heifner (changes requested):

libraries/chain/block_log.cpp
```diff
@@ -401,6 +413,15 @@ namespace eosio { namespace chain {
       FC_LOG_AND_RETHROW()
    }
 
+   void detail::block_log_impl::update_head(const signed_block_ptr& b) {
```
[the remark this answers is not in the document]
linh2931: Thanks.

tests/block_log_retain_blocks_test.py
```diff
@@ -0,0 +1,98 @@
...
+# the second for --block-log-retain-blocks 10,
+# the third for -block-log-retain-blocks not configured
+pnodes=1
+total_nodes=pnodes + 2
+
+try:
+    TestHelper.printSystemInfo("BEGIN")
+    cluster.setWalletMgr(walletMgr)
+
+    cluster.setWalletMgr(walletMgr)
```
heifner, line 50: Duplicate line.

heifner (approved):

tests/block_log_retain_blocks_test.py
linh2931: Thanks for careful reviewing! I am fixing it.

PR #647, opened 2022-07-12
[3.2] controller::extract_chain_id_from_db() does not properly handle dirty db

766C6164:
This change adds following logic:
- all exceptions  except system_error are propagated. system_error happens for example if database do not exist and needs to be created
- if chain_id cannot be retrieved empty variant returned, get replaced with find

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -3480,10 +3480,7 @@ std::optional<chain_id_type> controller::extract_chain_id_from_db( const path& s
       if (gpo==nullptr) return {};
 
       return gpo->chain_id;
-
-   } catch (std::system_error &) {   
-   
-   } 
+   } catch (std::system_error &) {} //  do not propagate db_error_code::not_found" for absent db, so it will be created 
 
    return {};
 }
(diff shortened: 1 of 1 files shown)


libraries/chain/controller.cpp
```diff
@@ -3476,11 +3476,14 @@ std::optional<chain_id_type> controller::extract_chain_id_from_db( const path& s
...
-      throw;
-   } catch( ... ) {
-   }
+      auto * gpo = db.find<global_property_object>();
+      if (gpo==nullptr) return {};
+
+      return gpo->chain_id;
+
+   } catch (std::system_error &) {   
+
```
heifner, line 3485: Can you add a comment about why we are ignoring `std::system_error`. Also please follow code style found in file.

heifner (changes requested):

libraries/chain/controller.cpp
766C6164: > Can you add a comment about why we are ignoring `std::system_error`. Also please follow code style found in file.

Sure, this was done to handle situation when there was no database and it needs to be created (db_error_code::not_found" exception). Not sure if the right way to plug it here or fix it differently.

heifner: Sorry, I meant to add a comment to the code.

766C6164: Ok, got it :)

heifner (approved):

Issue #651, opened 2022-07-12
deep-mind writes packed transaction instead of signed one on DTRX_OP PUSH_CREATE  


fschoell:
Looking at the deep-mind plugin, the version that has been integrated into eosio v2.1 writes signed transactions into the dmlog on `PUSH_CREATE` deferred transactions, see [here](https://github.com/EOSIO/eos/blob/release/2.1.x/libraries/chain/transaction_context.cpp#L695).

The current mandel seems to use the packed transaction instead though, see [here](https://github.com/eosnetworkfoundation/mandel/blob/main/libraries/chain/deep_mind.cpp#L225). This should be only used for `MODIFY_CREATE` while `PUSH_CREATE` should write the signed transaction.

heifner:
Resolved by #654

PR #654, opened 2022-07-12
[3.1] Use signed transaction for PUSH_CREATE action deep-mind logging.

oschwaldp-oci:
Resolves: https://github.com/eosnetworkfoundation/mandel/issues/651

Looking at the deep-mind plugin, the version that has been integrated into eosio v2.1 writes signed transactions into the dmlog on `PUSH_CREATE` deferred transactions, see [here](https://github.com/EOSIO/eos/blob/release/2.1.x/libraries/chain/transaction_context.cpp#L695).

The current mandel seems to use the packed transaction instead though, see [here](https://github.com/eosnetworkfoundation/mandel/blob/main/libraries/chain/deep_mind.cpp#L225). This should be only used for `MODIFY_CREATE` while `PUSH_CREATE` should write the signed transaction.

### DIFF
--- a/libraries/chain/deep_mind.cpp
+++ b/libraries/chain/deep_mind.cpp
@@ -225,6 +225,23 @@ namespace eosio::chain {
          ("trx", fc::to_hex(gto.packed_trx.data(), gto.packed_trx.size()))
       );
    }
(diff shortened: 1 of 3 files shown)


heifner (approved):

spoonincode (dismissed):
one sec

fschoell:
Great, seems to work. Thanks

PR #669, opened 2022-07-13
[3.1 -> main] Bring GH-651 and GH-646 into main

oschwaldp-oci:
[3.1 -> main] GH-651 Use signed transaction for PUSH_CREATE action deep-mind logging.
[3.1 -> main] GH-646 make fc::read_file_contents() check for errors instead of silently returning empty string


heifner (changes requested):
Looks like the wrong fc branch. This should be `main` of `fc`

oschwaldp-oci:
replaced by: https://github.com/eosnetworkfoundation/mandel/pull/670

Issue #639, opened 2022-07-11
populate through a mandel-dev.deb to mandel-contracts tests


spoonincode:
Currently mandel-contracts gets the native contract unit testing libraries & include files via the mandel CI storing a docker image of these assets on each mandel build.

This is fairly undesirable and sloppy, if anything because we're spamming github with container images that will never be used and live forever. Currently somewhere in the 1TB range I believe, and it'll just grow unbounded forever.

Instead, the procedure should be:

**For mandel repo**
* Build and package a mandel-dev.deb and leave it as an artifact in each workflow run
* Upon a release being minted, fetch the artifact matching the tagged release and bake an `experimental-binaries` package with it

**For mandel-contracts repo** it should allow configuration of either a tag, branch, or commit for its mandel-dev package to download and use. In general its process should be
* Try and find an `experimental-binaries` with the configured value, use that mandel-dev.deb if found
* Look for a mandel branch with the configured value, find its head, find the last workflow run corresponding to that commit, and use the mandel-dev.deb package artifact there
  * This won't work if the branch as a commit head that didn't run through CI. It's probably fine for something like a `release/3.1.x` branch though.
  * This will only be able to find the artifact if it's less than 90 days old. Not likely a problem as one would expect the branch to have been tagged in that time frame
* Look for a workflow run with the configured value as a commit hash, and use the mandel-dev.deb package artifact there
  * Same 90 days concern as above

spoonincode:
Will be resolved by eosnetworkfoundation/eos-system-contracts#39

PR #680, opened 2022-07-15
embed contracts required for native contract unit testing in to libtester

### LINKED ISSUE #642: investigate & solve something spooky in mandel-dev.deb
Integration with mandel-dev.deb + mandel-contracts + cdt + DUNE is showing some spooky behavior in mandel-dev.deb. What's going on?

spoonincode:
For native contract unit testing, libtester potentially needs to `setcode` a few bios contracts at run time on the chain driving unit tests. Previously, when libtester needed these contracts it would load them from the absolute path of which mandel was built. But after a `make dev-install` of mandel it's possible the build directory has been removed. And for the mandel-dev.deb package the build directory never even existed on the target.

For these few bios contracts bundle them inside of libtester so that there is no need to read them from disk. This allows libtester to access them after being installed from mandel-dev.deb, for example. I was attempting to minimize code changes in this PR; I think maybe some items should be rearranged further as commented on.

Resolves #642. Work on #639.

Picks up eosnetworkfoundation/mandel-fc#46 where I moved incbin.h in to fc (and updated to latest version) (also we're seeing more fc changes picked up in this PR because a few things are stacked up...)


libraries/testing/CMakeLists.txt
```diff
@@ -4,10 +4,13 @@ configure_file(contracts.hpp.in include/testing_contracts/contracts.hpp ESCAPE_Q
 add_library(eosio_testing_contracts INTERFACE)
 target_include_directories(eosio_testing_contracts INTERFACE ${CMAKE_CURRENT_BINARY_DIR}/include/testing_contracts)
 
+configure_file(contracts.cpp.in contracts.cpp ESCAPE_QUOTES)
```
spoonincode, line 7: There is a defect here. `contracts.cpp` depends on the various wasm/abi files it embeds and the build system is oblivious to that. Unfortunately attaching proper dependencies is going to require some reorganization due to these wasm/abi files being placed in to `CMAKE_BINARY_DIR` via `unittests/contracts`, which is currently executed after `libraries/testing`. I don't have a low touch fix for the reorg required to properly fix.

heifner (approved):

larryk85 (approved):

PR #690, opened 2022-07-18
Added FreeBSD build instructions

766C6164:
Added FreeBSD build instruction into docs. Its officially not supported, however it builds/runs/passes tests perfectly and instructions can be a useful for BSD users.

### DIFF
--- a/docs/00_install/01_build-from-source/index.md
+++ b/docs/00_install/01_build-from-source/index.md
@@ -94,31 +94,6 @@ make -j $(nproc) package
 After building you may remove the `$HOME/boost1.79` directory, or you may keep it around until next time building the software.
 </details>
 
-<details>
-  <summary>FreeBSD 13.1 Build Instructions</summary>
-
-Install required dependencies: 
-```
-pkg update && pkg install   \
-    git                     \
-    cmake                   \
-    curl                    \
-    boost-all               \
-    python3                 \
-    openssl                 \
-    llvm11                  \
-    pkgconf
-```
-and perform the build (please note that FreeBSD 13.1 comes with llvm13 by default so you should provide clang11 options to cmake):
-```
-git submodule update --init --recursive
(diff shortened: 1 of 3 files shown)


docs/00_install/01_build-from-source/index.md
```diff
@@ -94,6 +94,31 @@ make -j $(nproc) package
...
+    git                     \
+    cmake                   \
+    curl                    \
+    boost-all               \
+    python3                 \
+    openssl                 \
+    llvm11                  \
+    pkgconf
+```
+and perform the build (please note that FreeBSD 13.1 comes with llvm13 by default so you should provide clang11 options to cmake):
```
spoonincode, line 112: LLVM should actually not be required on freebsd; it's only required on Linux. The code that uses LLVM in mandel isn't clutched in on anything but Linux
https://github.com/eosnetworkfoundation/mandel/blob/b5132e5712ecf845abf3bf86f34fb87f99158338/CMakeLists.txt#L72-L76

Did you try just using the latest default clang that comes with freebsd? Historically clang12+ w/ GNU's libstdc++ (how typically configured on Linux) breaks due to still unfixed defect in EOS VM:
* https://github.com/EOSIO/eos-vm/pull/210 
But that's just coincidentally broken at clang12 -- it doesn't have anything to do with llvm12+ being unsupported on Linux.

macOS seems to build fine with the latest Xcode which makes me wonder if latest clang may work with LLVM's libc++ on freebsd?

766C6164: By default vanilla installation of FreeBSD 13.1 comes with clang 13 and compilation fails with exact issue you pointed out :) So unfortunately clang needs to be rolled back there. FreeBSD port "llvm11" is just the easiest and most trouble-free way to install working clang11...

spoonincode (approved):

PR #705, opened 2022-07-21
merge #648, #680 & #702 from 3.1 to main

spoonincode:
merge #648, #680 & #702 from 3.1.x to main.

Fixes some problems with "mandel-dev" package and building on boost 1.80


spoonincode:
blerg I forgot fc. will deal with this later

PR #710, opened 2022-07-21
[3.2] Add command line option --snapshot-to-json

### LINKED ISSUE #151: snapshot json support
https://github.com/EOSIO/eos/pull/11058 added option to convert snapshots to JSON.
The output format should be changed slightly from what is in that PR so that the resulting JSON can be more easily read. 
Along with 11058, should add the ability to read in the JSON. We have rapid JSON dependency already, so rather straightforward to add read of the JSON.

ClaytonCalabrese:
> Add new command line option --snapshot-to-json which reads given snapshot.bin and generates a JSON equivalent file.

Resolves https://github.com/eosnetworkfoundation/mandel/issues/151

### DIFF
--- a/libraries/chain/include/eosio/chain/snapshot.hpp
+++ b/libraries/chain/include/eosio/chain/snapshot.hpp
@@ -342,7 +342,22 @@ namespace eosio { namespace chain {
          std::streampos          header_pos;
          std::streampos          section_pos;
          uint64_t                row_count;
+   };
 
+   class ostream_json_snapshot_writer : public snapshot_writer {
+      public:
+         explicit ostream_json_snapshot_writer(std::ostream& snapshot);
+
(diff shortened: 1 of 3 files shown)


spoonincode (changes requested):
The linked issue mentions that this should not be part of nodeos, and that a separate tool should be created instead.

spoonincode:
Let's first decide if we want to jump on to an `eosio-snapshot` like application or do a more through integrated solution by removing `eosio-blocklog` and have a single `eosio-util`. If we just go with an `eosio-snapshot` it's pretty easy to bang that out unlike more broad changes.

ClaytonCalabrese:
closing due to migration to leap https://github.com/AntelopeIO/leap/pull/24

Issue #362, opened 2022-06-09
Remove gmp dependency


arhag:
https://github.com/eosnetworkfoundation/mandel/issues/182 added the dependency to gmp for the `mod_exp` host function as well as (indirectly) for the altbn128 host functions. These are currently dynamically linked due to license issues.

To reduce risk of possible shared library differences between nodes impacting consensus, it is safer to statically link the functionality provided by gmp into nodeos. However, due to licensing issues, gmp cannot be statically linked.

Instead, investigate an alternative library that can provide the same required functionality for the above mentioned host functions, e.g. libtommath, use that statically linked code instead of gmp, and remove the gmp dependency. 

The investigation of the alternative library should also involve benchmarking to ensure there isn't a significant performance degradation in the above mentioned host functions due to the change in library.

stephenpdeos:
Presently, our targeted solution to replace with tommath library has shown significant performance costs revealed in testing. This work was historically prioritized due to concerns that gmp future updates may disrupt the consensus process, but upon further discussion, the level of risk is supportive of cutting this from the 3.1 release with the acceptance that this will be a high priority body of work for the next immediate release.

stephenpdeos:
Closing out, will be managed in above listed issues that mention

PR #717, opened 2022-07-24
[3.1] replace gmp with tommath for mod_exp host function implementation

spoonincode:
This is work on #362 to remove the GMP dependencies from the new crypto functions. This replaces the GMP usage for `mod_exp` with libtommath instead.

Sadly this dings performance quite a bit: expect nearly a 5x slow down.

Needs eosnetworkfoundation/mandel-fc#48


spoonincode:
See comments on #362, we're not going to take this due to various reasons

PR #719, opened 2022-07-25
[3.2] Enhancement: Add unique error message for subjective billing in the transaction trace logs

766C6164:
Decision was made not to change exception, just a message, in particular:

Functions affected: validate_account_cpu_usage, validate_account_cpu_usage_estimate and checktime()

It is suggested that error messages with the condition of subjective_billed_us > 0 should be amended with following string:

" with a subjective cpu of (${subjective} us)"

for example, in validate_account_cpu_usage message should be updated as following:

Old:
"estimated CPU time (${billed} us) is not less than the maximum billable CPU time for the transaction (${billable} us)"

New:
"estimated CPU time (${billed} us) is not less than the maximum billable CPU time for the transaction (${billable} us) with a subjective cpu of (${subjective} us)"

Provisions has been made to avoid unneeded duplication of messages in a code.

### DIFF
--- a/libraries/chain/transaction_context.cpp
+++ b/libraries/chain/transaction_context.cpp
@@ -138,7 +138,7 @@ namespace eosio { namespace chain {
       initial_objective_duration_limit = objective_duration_limit;
 
       if( explicit_billed_cpu_time )
-         validate_cpu_usage_to_bill( billed_cpu_time_us, std::numeric_limits<int64_t>::max(), false ); // Fail early if the amount to be billed is too high
+         validate_cpu_usage_to_bill( billed_cpu_time_us, std::numeric_limits<int64_t>::max(), false, subjective_cpu_bill_us); // Fail early if the amount to be billed is too high
 
       // Record accounts to be billed for network and CPU usage
       if( control.is_builtin_activated(builtin_protocol_feature_t::only_bill_first_authorizer) ) {
@@ -204,7 +204,7 @@ namespace eosio { namespace chain {
          if( validate_account_cpu_limit > 0 )
             validate_account_cpu_limit -= EOS_PERCENT( validate_account_cpu_limit, 10 * config::percent_1 );
          if( validate_account_cpu_limit < 0 ) validate_account_cpu_limit = 0;
-         validate_account_cpu_usage_estimate( billed_cpu_time_us, validate_account_cpu_limit );
+         validate_account_cpu_usage_estimate( billed_cpu_time_us, validate_account_cpu_limit, subjective_cpu_bill_us );
       }
 
       eager_net_limit = (eager_net_limit/8)*8; // Round down to nearest multiple of word size (8 bytes) so check_net_usage can be efficient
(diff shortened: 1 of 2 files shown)


libraries/chain/transaction_context.cpp
```diff
@@ -496,17 +496,26 @@ namespace eosio { namespace chain {
             } else {
```
heifner, line 496: Just because greylist is in affect, doesn't mean it can't be limited by subjective cpu billing. Update the greylisted message as well.

libraries/chain/transaction_context.cpp
```diff
@@ -496,17 +496,26 @@ namespace eosio { namespace chain {
...
+                              tx_cpu_usage_exceeded,
+                              "billed CPU time (${billed} us) is greater than the maximum billable CPU time for the transaction (${billable} us) with a subjective cpu of (${subjective} us)",
+                              ("billed", billed_us)( "billable", cpu_limit )( "subjective", subjective_billed_us )
+                  );
+               }
+               else {
+                  EOS_ASSERT( billed_us <= cpu_limit,
+                              tx_cpu_usage_exceeded,
+                              "billed CPU time (${billed} us) is greater than the maximum billable CPU time for the transaction (${billable} us)",
+                              ("billed", billed_us)( "billable", cpu_limit )
```
heifner, line 510: Since this is also going to be used in the greylist case, I think simplifying it into one EOS_ASSERT would be cleaner.

libraries/chain/transaction_context.cpp
```diff
@@ -496,17 +496,26 @@ namespace eosio { namespace chain {
...
+                  );
+               }
             }
          }
       }
    }
```
heifner, line 517: I think we should also update the checktime() messages to include subjective billing as well.

heifner (changes requested):

heifner:
Can you update the PR description with example log messages.

libraries/chain/transaction_context.cpp
766C6164: Included these as well, thank you!

libraries/chain/transaction_context.cpp
766C6164: I tried to avoid duplication of same string constants over and over, hopefully without sacrificing readability too much...

libraries/chain/transaction_context.cpp
766C6164: done but not sure that the same message add-on should be used here

libraries/chain/transaction_context.cpp
```diff
@@ -487,26 +492,29 @@ namespace eosio { namespace chain {
...
-                           greylist_cpu_usage_exceeded,
-                           "billed CPU time (${billed} us) is greater than the maximum greylisted billable CPU time for the transaction (${billable} us)",
-                           ("billed", billed_us)( "billable", account_cpu_limit )
-               );
-            } else {
+            auto check_cpu_limit = [&](bool graylisted, bool subjective) {
+               std::string assert_msg =   "billed CPU time (${billed} us) is greater than the maximum";
+               assert_msg += graylisted ? " greylisted" : "";
+               assert_msg +=              " billable CPU time for the transaction (${billable} us)";
+               assert_msg += subjective ? " with a subjective cpu of (${subjective} us)" : "";
```
heifner, line 499: This `string` is created even when the `EOS_ASSERT` is not triggered. Please refactor so that time not wasted when not EOS_ASSERTing.

libraries/chain/transaction_context.cpp
heifner: I don't see the checktime() modification.

libraries/chain/transaction_context.cpp
heifner: I understand, it gets complicated when trying to avoid extra work and clean code.

libraries/chain/transaction_context.cpp
766C6164: Done, should be more efficient now. Thank you!

heifner (approved):

Issue #221, opened 2022-05-12
deep_mind in chain_plugin.cpp: allow named pipes


cc32d9:
it will be practical if deep-mind logs could be sent to a FIFO file instead of stdout. This requires one more configuration option and a small change in plugin initialization: it needs to check that the file exists and is a FIFO pipe, and set it as the logger output.

There's no performance impact, as FIFO is basically the same as stdout, but with a better flexibility. Also the deep-mind reader can disconnect and reconnect to the pipe, without stopping the nodeos.

linh2931:
After extensive discussions, it is decided to not proceed with short-term changes to deep-mind. Please see the details in the 2 PRs.

PR #723, opened 2022-07-25
[3.2] FIFO (named pipe) support for deep-mind logging

linh2931:
Resolve https://github.com/eosnetworkfoundation/mandel/issues/221.

A `nodeos` new option is added to support FIFO deep-mind logging:

```
  --deep-mind-fifo arg                     FIFO (named pipe) path for deep-mind
                                           logging. It is created either via
                                           mkfifo beforehand or dynamically. Only
                                           one of deep-mind and deep-mind-fifo can
                                           be enabled
```

Unit tests and integration tests are added.

https://github.com/eosnetworkfoundation/mandel-fc/pull/49 is the changes required in FC submodule.

### DIFF
--- a/plugins/chain_plugin/chain_plugin.cpp
+++ b/plugins/chain_plugin/chain_plugin.cpp
@@ -30,6 +30,7 @@
 
 #include <fc/io/json.hpp>
 #include <fc/variant.hpp>
+#include <fc/log/logger_config.hpp>
 #include <signal.h>
 #include <cstdlib>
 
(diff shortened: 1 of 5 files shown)


spoonincode:
I'd like to understand a little more about the use case of this and what the limitations/dangers are and if we are content with those.

The linked issue mentions
> Also the deep-mind reader can disconnect and reconnect to the pipe, without stopping the nodeos.

as a reason for this change, but this PR as is does not accommodate that from what I can see with a quick look over. If the consumer disconnects it appears impossible for it to sanely reconnect. Minimally it'll miss messages (nodeos won't wait for the consumer to reconnect, nor will it buffer messages), but it's also possible it'll reconnect mid-message and it'll have to be smart enough to "resync".

What this does allow is restarting nodeos without restarting the consumer. And maybe that's useful?

Also, whenever we add POSIXy stuff like this change contains, I'd like for us to take a swing at `#ifdef _WIN32`ing it. It doesn't have to be perfect and correct (as we have to real easy way to test) but it could help us out in the future.

linh2931:
Close this together with https://github.com/eosnetworkfoundation/mandel-fc/pull/49. Please see discussions there.

Issue #406, opened 2022-06-16
plugin_http_api_test improvement


linh2931:
https://github.com/eosnetworkfoundation/mandel/pull/387 fixes a race condition in activating a hardcoded list of protocol features by carefully arranging the list to get the test to pass. This needs to be revisited.

arhag:
We don't need to hard-code the protocol features and their digests in the Python file. Instead it can dynamically grab the list of supported protocol features using `get_supported_protocol_features` and then activate them.

stephenpdiesel:
Attempting to re-open and close again to resolve a GitHub bug. Please disregard.

Issue #664, opened 2022-07-13
Duplicate digest in expected_digest_list of plugin_http_api_test.py


linh2931:
While working on https://github.com/eosnetworkfoundation/mandel/issues/462, I noticed in `plugin_http_api_test.py`, `expected_digest_list` contains duplicate digest `4fca8bd82bbd181e714e283f83e1b45d95ca5af40fb89ad3977b653c448f78c2` (of `WEBAUTHN_KEY`). As a result, `BLOCKCHAIN_PARAMETERS`'s digest was missing. This is not a serious problem. It just causes `BLOCKCHAIN_PARAMETERS` is not verified in the test. No need to put into Rel 3.1.x.

arhag:
May be something that is automatically resolved by the improvements of https://github.com/eosnetworkfoundation/mandel/issues/406.

linh2931:
Fixed by https://github.com/eosnetworkfoundation/mandel/pull/725.

PR #725, opened 2022-07-26
[3.2] Replace hardcoded protocol feature names and digests with dynamically retrieved names and digests

linh2931:
Resolve https://github.com/eosnetworkfoundation/mandel/issues/406 and https://github.com/eosnetworkfoundation/mandel/issues/664.

Currently `plugin_http_api_test.py` hardcoded protocol feature names and digests. The test has to be updated when old protocol features change or removed, and when new protocol features are added. Another problem is it is error prone to hardcode digests, as reported by https://github.com/eosnetworkfoundation/mandel/issues/664 -- duplicated digests are coded.

This PR dynamically retrieves builtin protocol features from the chain.

It is noted during review the number of protocol features was hardcoded too. Corrected it.


tests/plugin_http_api_test.py
```diff
@@ -113,6 +113,7 @@ def test_ChainApi(self) :
         self.activateAllBuiltinProtocolFeatures()
         allProtocolFeatures = self.nodeos.getSupportedProtocolFeatures()
         allFeatureDigests = [d['feature_digest'] for d in allProtocolFeatures]
+        allFeatureCodenames = [s['specification'][0]['value'] for s in allProtocolFeatures]
```
spoonincode, line 116: It would be nice if we could make sure we're looking at `name == builtin_feature_codename` before assuming it's the `value` we want.. dunno if complexity is worth it tho. Fine as is imo

spoonincode (approved):

tests/plugin_http_api_test.py
linh2931: Thanks for the comments. I will add the check.

PR #732, opened 2022-07-27
[3.2] Implement JSON Snapshot Reader

ClaytonCalabrese:
https://github.com/eosnetworkfoundation/mandel/pull/710 ports a way to convert a snapshot to json. This PR enhances the `--snapshot` option to support reading a snapshot in JSON format if the snapshot has `.json` extension.

Resolves: https://github.com/eosnetworkfoundation/mandel/issues/771


libraries/chain/snapshot.cpp
```diff
@@ -386,6 +337,82 @@ void istream_snapshot_reader::return_to_header() {
...
 }
 
+istream_json_snapshot_reader::istream_json_snapshot_reader(std::istream& snapshot)
+   :snapshot(snapshot)
+   ,header_pos(snapshot.tellg())
+   ,num_rows(0)
+   ,cur_row(0)
+{
+   std::stringstream my_buffer;
+   my_buffer << snapshot.rdbuf();
```
spoonincode, line 347: This looks like it reads the entire file in to memory? How much memory does this take for something like an EOS or WAX snapshot? It almost seems like that's going to be impractical.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -987,6 +989,52 @@ void chain_plugin::plugin_initialize(const variables_map& options) {
...
+         // used for validation code below
+         auto infile = std::ifstream(my->json_snapshot_path->generic_string(), (std::ios::in));
+         istream_json_snapshot_reader reader(infile);
+         reader.validate();
+         chain_id = controller::extract_chain_id(reader);
+         infile.close();
+
+         EOS_ASSERT( options.count( "genesis-timestamp" ) == 0,
+                 plugin_config_exception,
+                 "--snapshot is incompatible with --genesis-timestamp as the snapshot contains genesis information");
```
spoonincode, line 1007: `--json-snapshot` not `--snapshot`. But more generally this looks like 40 lines copy pasted (almost?) verbatim. Is there a better way?

spoonincode:
Have we (manually) tested that we can load something like a WAX snapshot in JSON format with this?

ClaytonCalabrese:
> Have we (manually) tested that we can load something like a WAX snapshot in JSON format with this?
I doubled my WSL max storage yesterday and attempted to run WAX but the snapshot-to-json ran my VM out of space at this row:
`,"row_288346839":{"primary_key":"1099801935006","payer":"farmersworld","value":"nrxNEQABAACQ4qUcXyWvWQAAAAAAHCnNbh`

I can allocate some more space as well as delete some stuff to try and fit all of wax's JSON Snapshot tonight, but I have no idea how much space the full thing is going to take to create.

libraries/chain/include/eosio/chain/snapshot.hpp
```diff
@@ -5,6 +5,7 @@
 #include <fc/variant_object.hpp>
 #include <boost/core/demangle.hpp>
 #include <ostream>
+#include "../rapidjson/include/rapidjson/document.h"
```
spoonincode, line 8: it really should be added to cmake as an INTERFACE library and properly set up so only need to `#include <rapidjson/document.h>`

spoonincode:
I think only one version of rapidjson should be in the global namespace so we avoid any ODR wonkiness . Right now, I believe abieos includes it in the global namespace too.

libraries/chain/CMakeLists.txt
```diff
@@ -131,6 +131,7 @@ target_link_libraries( eosio_chain PUBLIC fc chainbase Logging IR WAST WASM Runt
 target_include_directories( eosio_chain
                             PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}/include" "${CMAKE_CURRENT_BINARY_DIR}/include"
                                    "${CMAKE_CURRENT_SOURCE_DIR}/../wasm-jit/Include"
+                                   "${CMAKE_CURRENT_SOURCE_DIR}/../rapidjson/include"
```
spoonincode, line 134: This really needs to be an INTERFACE library

heifner: Can you provide an example of what that would look like. I didn't see an example in our code.

spoonincode: something like (I'm just winging this so may not be fully correct)
```
add_library(rapidjson INTERFACE)
target_include_directories(rapidjson INTERFACE rapidjson/include)
target_compile_definitions(rapidjson INTERFACE -DRAPIDJSON_NAMESPACE=eosio_rapidjson)
```

then wherever you want to use it just
```
target_link_libraries(eosio_chain rapidjson)
```

spoonincode: I'm not sure if making the RAPIDJSON_NAMESPACE thing part of the interface is a good idea or not though.

ClaytonCalabrese:
closing due to migration to leap https://github.com/AntelopeIO/leap/pull/25

PR #735, opened 2022-07-31
Replace cleos local abi file test with test harness version.

jgiszczak:
Replace the cleos local ABI file test with a version implemented using the test harness and actually execute it.  The previous version was never executed by the test file.

### DIFF
--- a/tests/cli_test.py
+++ b/tests/cli_test.py
@@ -357,7 +357,7 @@ def abi_file_with_nodeos_test():
         os.mkdir(data_dir)
         walletMgr = WalletMgr(True)
         walletMgr.launch()
-        node = Node('localhost', 8888, nodeId, cmd="./programs/nodeos/nodeos -e -p eosio --plugin eosio::trace_api_plugin --trace-no-abis --plugin eosio::producer_plugin --plugin eosio::producer_api_plugin --plugin eosio::chain_api_plugin --plugin eosio::chain_plugin --plugin eosio::http_plugin --access-control-allow-origin=* --http-validate-host=false " + "--data-dir " + data_dir + " --config-dir " + data_dir, walletMgr=walletMgr)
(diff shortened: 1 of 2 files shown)


tests/cli_test.py
```diff
@@ -360,33 +344,87 @@ def cleos_abi_file_test():
...
+                assert False
+            tries -= 1
+            time.sleep(2)
+        nodeId = 'bios'
+        data_dir = Utils.getNodeDataDir(nodeId)
+        assert not os.path.exists(data_dir), 'data_dir exists'
+        os.mkdir(data_dir)
+        walletMgr = WalletMgr(True)
+        walletMgr.launch()
+        node = Node('localhost', 8888, nodeId, cmd="./programs/nodeos/nodeos -e -p eosio --plugin eosio::trace_api_plugin --trace-no-abis --plugin eosio::producer_plugin --plugin eosio::producer_api_plugin --plugin eosio::chain_api_plugin --plugin eosio::chain_plugin --plugin eosio::http_plugin --access-control-allow-origin=* --http-validate-host=false " + "--data-dir " + data_dir + " --config-dir " + data_dir, walletMgr=walletMgr)
```
spoonincode, line 360: I think this may need `--resource-monitor-not-shutdown-on-threshold-exceeded` & its own `config-dir`

tests/cli_test.py
```diff
@@ -360,33 +344,87 @@ def cleos_abi_file_test():
...
+            Utils.Print("Test failed.")
+        if node:
+            if not node.killed:
+                if node.pid:
+                    os.kill(node.pid, signal.SIGKILL)
+        if testSuccessful:
+            Utils.Print("Cleanup nodeos data.")
+            shutil.rmtree(data_dir)
+
+        '''
```
spoonincode, line 412: any reason to comment this out instead of removing it?

jgiszczak: Forgot this was here after removing other commented code.  Removed.

tests/cli_test.py
jgiszczak: Added resource monitor flag.  It has a specified `config-dir`.

spoonincode (approved):

PR #742, opened 2022-08-03
[3.1 -> main] Add subjective limit to mod_exp host function

### LINKED ISSUE #361: Add subjective limit for mod_exp
https://github.com/eosnetworkfoundation/mandel/issues/182 adds a host function `mod_exp` for modular exponentiation. There are currently no limits on that. It may be necessary to enforce some limits to prevent the actual termination of a transaction from being delayed too much beyond the intended deadline timeout.

Benchmarking should be done to determine what are appropriate permissive limits that prevent a contract from creating too much of a delay (less than a millisecond?) within the process of aborting the transaction. Those limits should then be enforced in the `mod_exp` host function as a subjective exception (meaning it would be enforced at the BPs and not actually part of the protocol).

arhag:
Brings #739 to main.

Resolves #361.

### DIFF
--- a/libraries/chain/webassembly/crypto.cpp
+++ b/libraries/chain/webassembly/crypto.cpp
@@ -158,6 +158,14 @@ namespace eosio { namespace chain { namespace webassembly {
                               span<const char> exp,
                               span<const char> modulus,
                               span<char> out) const {
+      if (context.control.is_producing_block()) {
+         static constexpr int byte_size_limit = 256; // Allow up to 2048-bit values for base, exp, and modulus.
+
+         if (std::max(std::max(base.size(), exp.size()), modulus.size()) > byte_size_limit) {
+            EOS_THROW(subjective_block_production_exception, "Bit size too large for values passed into mod_exp");
+         }
+      }
+
       bytes bbase(base.data(), base.data() + base.size());
       bytes bexp(exp.data(), exp.data() + exp.size());
       bytes bmod(modulus.data(), modulus.data() + modulus.size());
--- a/unittests/crypto_primitives_tests.cpp
+++ b/unittests/crypto_primitives_tests.cpp
@@ -386,6 +386,7 @@ BOOST_AUTO_TEST_CASE( modexp_test ) { try {
          return_code::failure,
          "",
       },
+
    };
 
    for(const auto& test : tests) {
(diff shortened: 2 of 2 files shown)


larryk85 (approved):

libraries/chain/webassembly/crypto.cpp
```diff
@@ -158,6 +158,14 @@ namespace eosio { namespace chain { namespace webassembly {
                               span<const char> exp,
                               span<const char> modulus,
                               span<char> out) const {
+      if (context.control.is_producing_block()) {
+         static constexpr int byte_size_limit = 256; // Allow up to 2048-bit values for base, exp, and modulus.
```
smlu, line 162: How was the 2048 bit constraint chosen?
It's normal to have 4096 bit modulus nowadays, especially with root certificates.
If there isn't big performance hit (CPU, RAM) I'd suggest to lift this constraint to 512.

arhag: See reasoning in this comment within the code:
https://github.com/eosnetworkfoundation/mandel/blob/6514a7d6379000588c3923528e4da559c5a40c60/unittests/crypto_primitives_tests.cpp#L415-L423

The 5 ms is somewhat arbitrary. The idea is to not have it be too large of a fraction of the time it takes to deliver a complete block to the next producer to avoid missing blocks. If we assume a typical window of 100 ms time given based on configuration in the producer plugin, than bounding it to 5 ms eats up less than 5% of that time reserved to complete the block, sign it, and then start working on the next one (if not at the end of the round for that producer) or deliver over the network to the next producer in the schedule (if at the end of the round for that producer in which case they typically have more time available as well than 100 ms based on default configurations).

While bumping up to 4096-bit is still well within the typical 30 ms deadline limit, it ends up eating a significant portion of that 100 ms delivery window (around 20% instead of 5%). So we weren't very comfortable pushing it that far just yet.

The proper solution is to remove these arbitrary bit-size limits and just use deadline time as the only practical bound on what calls to `mod_exp` are allowed. Unfortunately, that requires plumbing the deadline timer into the inner loops of the modular exponentiation implementation which is something we intend to do (https://github.com/eosnetworkfoundation/mandel/issues/738) but is not something we are able to do for the 3.1.0 release.

arhag: I'm now going with a more flexible constraint on the bit sizes which should allow for 4096-bit RSA to work.

See https://github.com/eosnetworkfoundation/mandel/issues/747 and https://github.com/eosnetworkfoundation/mandel/pull/749.

smlu: Oh that's great change!

Btw do you maybe have any metrics how long does it take to perform mod. exp. for 4096 bit modulus with current implementation? Hitting 5ms or more for 4096 bit  sounds a bit too much. (Sorry I'm unable to perform the benchmark myself at the moment).

Just for reference, I made dumb benchmark for [powm](https://github.com/ZeroPass/eosio.ck/blob/master/include/eosiock/c/powm.h) implementation from `eosio.ck` (originally from u-boot) and compiled with clang [64bit -O3 -stdlib=libc++ -std=c++2a].
 I get ~500us for single mod. exp. operation over 4096 bit modulus & base and on average ~468us per operation running in the loop 1M times (~2100 oper./s). Test vector was taken from Google's [project Wycheproof](https://github.com/google/wycheproof/blob/master/testvectors/rsa_pss_4096_sha512_mgf1_32_test.json#L14-L30).

arhag: On my machine, `mod_exp` with 4096-bit modulus and base, and a 32-bit exponent took less than 200 microseconds.

PR #743, opened 2022-08-03
[3.2] grab bag of warning fixes

spoonincode:
Most of these are to fix `comparison of integer expressions of different signedness` by replacing `int` with `size_t` for comparison against `vector<>::size()` unless otherwise noted

### DIFF
--- a/unittests/eosio.token_tests.cpp
+++ b/unittests/eosio.token_tests.cpp
@@ -152,7 +152,7 @@ BOOST_FIXTURE_TEST_CASE( create_max_supply, eosio_token_tester ) try {
    share_type amount = 4611686018427387904;
    static_assert(sizeof(share_type) <= sizeof(asset), "asset changed so test is no longer valid");
    static_assert(std::is_trivially_copyable<asset>::value, "asset is not trivially copyable");
-   memcpy((char*)&max, &amount, sizeof(share_type)); // hack in an invalid amount
+   memcpy(&max, &amount, sizeof(share_type)); // hack in an invalid amount
 
    BOOST_CHECK_EXCEPTION( create( "alice"_n, max) , asset_type_exception, [](const asset_type_exception& e) {
       return expect_assert_message(e, "magnitude of asset amount must be less than 2^62");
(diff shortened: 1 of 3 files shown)


unittests/eosio.token_tests.cpp
```diff
@@ -177,7 +177,7 @@ BOOST_FIXTURE_TEST_CASE( create_max_decimals, eosio_token_tester ) try {
    share_type amount = 0x8ac7230489e80000L;
    static_assert(sizeof(share_type) <= sizeof(asset), "asset changed so test is no longer valid");
    static_assert(std::is_trivially_copyable<asset>::value, "asset is not trivially copyable");
-   memcpy(&max, &amount, sizeof(share_type)); // hack in an invalid amount
+   memcpy((char*)&max, &amount, sizeof(share_type)); // hack in an invalid amount
```
spoonincode, line 180: These were `copying an object of non-trivial type` warnings. Gnarly fix but, well, comment says it's a hack :wink:

unittests/state_history_tests.cpp
```diff
@@ -490,7 +490,7 @@ BOOST_AUTO_TEST_CASE(test_deltas_resources_history) {
 
       std::multiset<std::string> expected_contract_row_table_names {"abihash", "abihash", "hashobjs", "hashobjs", "hashobjs", "numobjs", "numobjs", "numobjs"};
 
-      std::multiset<uint64_t> expected_contract_row_table_primary_keys {6138663577826885632,14605619288908759040, 0, 1 ,2, 0, 1, 2};
+      std::multiset<uint64_t> expected_contract_row_table_primary_keys {6138663577826885632U,14605619288908759040U, 0, 1 ,2, 0, 1, 2};
```
spoonincode, line 493: `integer constant is so large that it is unsigned`

unittests/eosio.token_tests.cpp
spoonincode: I removed this fix for now, so the other clearly correct ones can be fast pathed in

heifner (approved):

PR #749, opened 2022-08-05
[3.1] Use more flexible subjective limits on mod_exp.

arhag:
A new approach to constraining the bit sizes of the arguments passed into mod_exp is taken which allows for larger bit sizes for the base and modulus arguments when the exponent is small.

This now allows for mod_exp to be used for 4096-bit RSA.

### DIFF
--- a/libraries/chain/webassembly/crypto.cpp
+++ b/libraries/chain/webassembly/crypto.cpp
@@ -176,7 +176,7 @@ namespace eosio { namespace chain { namespace webassembly {
                       "mod_exp restriction: exponent bit size cannot exceed bit size of either base or modulus");
          }
 
-         static constexpr uint64_t bit_calc_limit = 101;
+         static constexpr uint64_t bit_calc_limit = 106;
 
          uint64_t bit_calc = 5 * ceil_log2(exp.size()) + 8 * ceil_log2(base_modulus_size);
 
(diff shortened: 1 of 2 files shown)


libraries/chain/webassembly/crypto.cpp
```diff
@@ -159,10 +169,20 @@ namespace eosio { namespace chain { namespace webassembly {
...
       if (context.control.is_producing_block()) {
-         static constexpr int byte_size_limit = 256; // Allow up to 2048-bit values for base, exp, and modulus.
+         unsigned int base_modulus_size = std::max(base.size(), modulus.size());
+
+         if (base_modulus_size < exp.size()) {
+            EOS_THROW(subjective_block_production_exception, 
+                      "mod_exp restriction: exponent bit size cannot exceed bit size of either base or modulus");
+         }
+
+         static constexpr uint64_t bit_calc_limit = 101;
```
larryk85, line 179: Where did 101 come from? I think we still want to support 2048 bit exponent if base or modulus is still max 2048 bits? In that case it would equal 104.

smlu: If exponent is small enough (which usually is for RSA crypto) 2048+ bit modulus can be used. 
e.g standard RSA exponent 0x010001 with 2048 bit modulus gives 74 bits with above formula.

smlu: Ok yeah you're right, for backward compatibility with previous semantics  constraint should be `104` to get in all 3 ints at 2048 bits.

arhag: I updated it from 101 to 106. See the fc submodule comments for reasoning.

larryk85 (approved):

Issue #285, opened 2022-05-20
Investigate if Cluster requires Trace API Plugin


oschwaldp-oci:
Investigate whether Cluster requires Trace API Plugin to be enabled now to work at all. If so, we should just have Cluster automatically add this if not already added instead of requiring most tests to add the launch config params:

`traceNodeosArgs = " --plugin eosio::trace_api_plugin --trace-no-abis "`

Initial investigation reveals that the requirement stems from whether `Cluster.py` `launch` 's function will use `bootstrap` or `bios_bootstrap`.

If `bootstrap` will be called which in turn goes into `Node.py` and calls `publishContract` with `waitForTransBlock=True` and eventually calls `getBlockNumByTransId` which requires a call to `getTransaction` and reaches out to cleos with a `get transaction_trace` command, which requires the Trace API.

If `bootstrap` is not used, in favor of conditions where `bios_bootstrap` is used, it may not require the Trace API plugin.

PR #750, opened 2022-08-05
[3.2] move trace api plugin activation out of tests and into cluster

ClaytonCalabrese:
> Investigate whether Cluster requires Trace API Plugin to be enabled now to work at all. If so, we should just have Cluster automatically add this if not already added instead of requiring most tests to add the launch config params:

> 
> traceNodeosArgs = " --plugin eosio::trace_api_plugin --trace-no-abis "
> 
> Initial investigation reveals that the requirement stems from whether Cluster.py launch 's function will use bootstrap or bios_bootstrap.
> 
> If bootstrap will be called which in turn goes into Node.py and calls publishContract with waitForTransBlock=True and eventually calls getBlockNumByTransId which requires a call to getTransaction and reaches out to cleos with a get transaction_trace command, which requires the Trace API.
> 
> If bootstrap is not used, in favor of conditions where bios_bootstrap is used, it may not require the Trace API plugin.
> 

Due to some updates to mandel, trace_api_plugin was needed across a wide number of tests. Previously this was done manually in each test by passing it to `extraNodeosArgs`. Rather than doing it that way, it is cleaner to activate in cluster, defaulting to `--trace-no-abis` unless the less common `--trace-rpc-abi` is passed in from a test.

Resolves https://github.com/eosnetworkfoundation/mandel/issues/285


tests/large-lib-test.py
```diff
@@ -68,15 +68,12 @@ def relaunchNode(node: Node, chainArg="", skipGenesis=True, relaunchAssertMessag
...
-    traceNodeosArgs=" --plugin eosio::trace_api_plugin --trace-no-abis "
     if cluster.launch(
             pnodes=pnodes,
             totalNodes=total_nodes,
             totalProducers=1,
             useBiosBootFile=False,
-            topo="mesh",
-            specificExtraNodeosArgs=specificExtraNodeosArgs,
-            extraNodeosArgs=traceNodeosArgs) is False:
+            topo="mesh") is False:
```
heifner, line 76: `specificExtraNodeosArgs` should be added back in.

heifner (changes requested):

heifner (approved):

PR #799, opened 2022-08-09
[3.2] benchmarking framework and its use in benchmarking crypto primitives

### LINKED ISSUE #780: Spike: research and recommend altbn128 host function support solution
We need to evaluate the best path forward to support altbn128 host functions via either Rust or Go.

Benchmarking for both solutions should be part of this investigation.

### LINKED ISSUE #786: Benchmark of crypto functions as separate executable
https://github.com/eosnetworkfoundation/mandel/issues/361 added a test case with does benchmarking for modular exponentiation. That should really be moved into a separate program for the purpose of benchmarking. Similarly, we should include benchmarking of other crypto functions:

- alt_bn128 math operations
- Signing and key recovery/validation for R1, K1, WebAuthn elliptic curves; 
- hash functions exposed via host functions, e.g. SHA256, (benchmark both small and large sizes of messages to hash).

linh2931:
Resolve https://github.com/eosnetworkfoundation/mandel/issues/786 and partially resolve https://github.com/eosnetworkfoundation/mandel/issues/780.

Implement a benchmarking framework to help writing benchmarking code by reusing common measuring functions and unifying reports. After the PR is approved, the framework can be used to benchmark other functions.

### For end users

Below are a few examples

**Help**

```
./benchmark -h
benchmark command line options:
  -m [ --module ] arg       the module to be benchmarked; if this option is not
                            present, all modules are benchmarked.
  -l [ --list ]             list of supported modules
  -r [ --runs ] arg (=1000) the number of runs per function
  -h [ --help ]             execute the benchmarked functions for the specified
                            number of times, and report average, minimum, and
                            maximum time in nanoseconds
```

**Benchmarking All Modules (functions)**

```
./benchmark
function                      runs               avg              min              max

alt_bn_128:
alt_bn128_add                 1,000         6,467 ns         3,727 ns       113,625 ns
alt_bn128_mul                 1,000       251,105 ns       235,337 ns       395,856 ns
alt_bn128_pair (1 pair)       1,000     6,405,913 ns     6,354,466 ns     8,265,015 ns
alt_bn128_pair (10 pairs)     1,000    25,379,638 ns    25,236,388 ns    36,013,377 ns

modexp:
1024-bit width                1,000       389,252 ns       388,298 ns       396,263 ns
2048-bit width                1,000     2,914,494 ns     2,910,697 ns     2,999,909 ns
```

**Benchmarking Particular Module with Specific Number of Runs**

```
/benchmark -m modexp -r 500
function                      runs               avg              min              max

modexp:
1024-bit width                500         441,057 ns       388,091 ns       818,064 ns
2048-bit width                500       2,914,402 ns     2,911,318 ns     3,001,094 ns
```

### For developers

Please see `alt_bn_128.cpp` and `modexp.cpp `for examples.

The general approach to adding code to benchmark a set of functions is

1.  Implement a function where for each function to be benchmarked, prepare input, define a lamda to call the function, and call common `benchmarking`.
2. Add an entry to `modules` in `benchmark.cpp`.


heifner:
Should we consider using https://github.com/google/benchmark which has built in support for preventing the compiler from optimizing out what you are trying to benchmark as well as other useful features for micro-benchmarking.

linh2931:
> 
The Google microbenchmarking library looks great. It has lots of features and more code. Just wondering if it is brought in, would it increase dependencies and build time?

stephenpdeos:
We will not move forward with Google benchmarking framework at this time, because it doesn't seem to have enough value add compared to current code used.

benchmark/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+file(GLOB BENCHMARK "*.cpp")
+add_executable( benchmark ${BENCHMARK} )
+
+target_link_libraries( benchmark fc Boost::program_options )
+target_include_directories( benchmark PUBLIC
+                            ${CMAKE_CURRENT_SOURCE_DIR}
```
spoonincode, line 6: This will need to be wrapped in quotes to handle spaces. You may also be able to just put a `.` here since paths are relative to `CMAKE_CURRENT_SOURCE_DIR` anyways. Or just forego it all together and `include "benchmark.hpp"` but that's typically not the pattern we use.

spoonincode:
I wonder if there should be a machine-readable output (like JSON) to make it easier to ingest in to some reporting framework.

benchmark/benchmark.cpp
```diff
@@ -0,0 +1,73 @@
...
+      << std::setw(time_width) << min << std::setw(ns_width) << " ns"
+      << std::setw(time_width) << max << std::setw(ns_width) << " ns"
+      << std::endl;
+}
+
+void benchmarking(std::string name, const std::function<void()>& func) {
+   uint64_t total {0}, min {std::numeric_limits<uint64_t>::max()}, max {0};
+
+   for (auto i = 0U; i < num_runs; ++i) {
+      auto start_time = std::chrono::steady_clock::now();
```
spoonincode, line 60: It feels like `high_resolution_clock` is what we would want when making tiny measurements like this, but it's not a monotonic clock on libstdc++ :confused:

benchmark/CMakeLists.txt
linh2931: Thanks for the suggestions.

linh2931:
> 

I will add a -j option for JSON format output.

benchmark/benchmark.cpp
larryk85: I don't think it matters for this use case if the clock is not monotonic. It's only a relative number that is just needed for orders of magnitudes and comparisons. So, I agree that this should be high_resolution_clock.

larryk85:
> I wonder if there should be a machine-readable output (like JSON) to make it easier to ingest in to some reporting framework.

That would be nice to have, but maybe we can list that as a phase 2 item for after this is merged and done.

benchmark/benchmark.cpp
linh2931: Thanks. Will change to use high resolution clock.

swatanabe: Only needing relative times is exactly the use for monotonic clocks.  It is probably okay to use a non-monotonic clock, but that's because the small chance of producing wild results if the system clock happens to change while running the benchmark is a manageable problem.

linh2931:
Close this as it is being ported over to Leap as https://github.com/AntelopeIO/leap/pull/17.

PR #817, opened 2022-08-15
Moved to using a sync socket implementation

ndcgundlach:
Changed to code to not use net plugin code except the buffer building for the transaction provider.

### DIFF
--- a/tests/trx_generator/trx_provider.cpp
+++ b/tests/trx_generator/trx_provider.cpp
@@ -1,19 +1,10 @@
 #include <trx_provider.hpp>
 
-#include <fc/network/message_buffer.hpp>
 #include <fc/network/ip.hpp>
-#include <fc/io/json.hpp>
 #include <fc/io/raw.hpp>
 #include <fc/log/appender.hpp>
-#include <fc/log/logger_config.hpp>
-#include <fc/log/trace.hpp>
-#include <fc/reflect/variant.hpp>
-#include <fc/crypto/rand.hpp>
-#include <fc/exception/exception.hpp>
 
 #include <boost/asio/ip/tcp.hpp>
-#include <boost/asio/ip/host_name.hpp>
-#include <boost/asio/steady_timer.hpp>
 
 using std::string;
 using std::vector;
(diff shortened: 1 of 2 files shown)


tests/trx_generator/trx_provider.cpp
```diff
@@ -20,17 +20,38 @@ using std::vector;
...
+   static send_buffer_type create_send_buffer( const chain::packed_transaction& m ) {
+      const uint32_t payload_size = fc::raw::pack_size( m );
 
+      const char* const header = reinterpret_cast<const char* const>(&payload_size); // avoid variable size encoding of uint32_t
+      const size_t buffer_size = message_header_size + payload_size;
+
+      auto send_buffer = std::make_shared<vector<char>>(buffer_size);
+      fc::datastream<char*> ds( send_buffer->data(), buffer_size);
+      ds.write( header, message_header_size );
+      fc::raw::pack( ds, fc::unsigned_int(8));
```
heifner, line 46: Add comment that `8` is the `which` of the variant for packed_transaction

tests/trx_generator/trx_provider.cpp
```diff
@@ -20,17 +20,38 @@ using std::vector;
...
+      fc::raw::pack( ds, m );
+
+      return send_buffer;
    }
 
    void p2p_connection::send_transaction(const chain::packed_transaction& trx) {
-
+      send_buffer_type msg = create_send_buffer(trx);
```
heifner, line 53: May need to separate the creation of the buffer from the send in the future as this creates a bit of time between each send. Not sure that extra time will ever matter.

tests/trx_generator/trx_provider.cpp
```diff
@@ -20,17 +20,38 @@ using std::vector;
...
-
+      p2p_socket.connect( tcp::endpoint( boost::asio::ip::address::from_string(_peer_endpoint), 8090));
+      boost::system::error_code error;
    }
 
    void p2p_connection::disconnect() {
+      p2p_socket.close();
+   }
+
+   constexpr auto     message_header_size = sizeof(uint32_t);
```
oschwaldp-oci, line 35: Could you clean up spacing here?

tests/trx_generator/trx_provider.cpp
ndcgundlach: It looks like I would need a max pack_size for packed_transaction to avoid allocating the char vector.  I would still have to pack the data into the buffer.  I would note that this function packs the packed transaction twice.

oschwaldp-oci (approved):

PR #821, opened 2022-08-16
Validate Number of Trx in Performance Harness Match Nodeos Log

ClaytonCalabrese:
Adds validation to check the number of expected transactions matches the number of transactions logged in nodeos.
Merges in the dev branch and fixes some bugs regarding rejected transactions.
Blocks on empty blocks to enter and leave test.

### DIFF
--- a/tests/performance_tests/performance_test_basic.py
+++ b/tests/performance_tests/performance_test_basic.py
@@ -18,17 +18,18 @@ Print = Utils.Print
 errorExit = Utils.errorExit
 cmdError = Utils.cmdError
 relaunchTimeout = 30
+emptyBlockGoal = 5
 
 def waitForEmptyBlocks(node1):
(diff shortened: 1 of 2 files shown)


tests/performance_tests/performance_test_basic.py
```diff
@@ -18,6 +19,28 @@
 errorExit = Utils.errorExit
 cmdError = Utils.cmdError
 relaunchTimeout = 30
+emptyBlockGoal = 5
+
+def waitForEmptyBlocks(node1):
```
oschwaldp-oci, line 24: we should probably make this generic `node` not specific since the method could be used on any node, and we just happen to call it on `node1`

oschwaldp-oci (approved):
Nothing to block merging, just a comment or two.

